PR #1, opened 2018-10-01
Make context header consistent with other projects

gunnihinn:
Go-carbon expects to see either X-CTX-CarbonAPI-UUID or
X-CTX-CarbonZipper-UUID, so let's give it one of those.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -11,7 +11,7 @@ import (
 type key int
 
 const (
-	ctxHeaderUUID = "X-CTX-Carbon-UUID"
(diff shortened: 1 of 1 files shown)


gunnihinn:
> Since you can build both carbonapi and carbonzipper shouldn't it be changed depending on type of application one's running?

No?

This just controls what HTTP header we use to communicate the request
UUID to go-carbon. The user doesn't care, and each user presumably knows
whether they told carbonapi or carbonzipper to talk to go-carbon, so
there shouldn't be new information there.

PR #4, opened 2018-10-04
Register limiter use metric with graphite

gksinghjsr:
Also, return the current host value, instead of a map


gksinghjsr:
This will only affect api, and not zipper, hence resulting in a metric each for the carbonapi fe boxes.

gksinghjsr:
Zipper stores the limiter count for each backend stores, and if implemented there, it would result in an explosion.

PR #12, opened 2018-10-19
Gysingh/fix zipper python

gksinghjsr:
Extracting api and zipper into their separate packages had a fallout wherein we no longer had access to the global config from other packages. This is an effort to inject the config into the handlers (as recievers).


gksinghjsr:
The branch name is a little misleading as the problem was first spotted while running python graphite with zipper.

app/carbonapi/api.go
```diff
@@ -40,26 +41,17 @@ import (
 // BuildVersion is provided to be overridden at build time. Eg. go build -ldflags -X 'main.BuildVersion=...'
 var BuildVersion string
 
-var config = struct {
-	cfg.API
```
gunnihinn: Was there a problem with embedding the `cfg.API` struct?

gksinghjsr: The EnvConfig struct embeds the cfg.API struct and some other configs as well, similar to the previous configs for api and zipper. The zipper EnvConfig wraps the cfg.Zipper and list of backends.

gksinghjsr: @gunnihinn do you have any concerns regarding this?

app/zipper/http_handlers.go
```diff
@@ -147,8 +147,10 @@ func findHandler(w http.ResponseWriter, req *http.Request) {
 	)
 
 	request := types.NewFindRequest(originalQuery)
-	bs := backend.Filter(backends, []string{originalQuery})
+	bs := backend.Filter(app.backends, []string{originalQuery})
 	metrics, err := backend.Finds(ctx, bs, request)
+	fmt.Println(metrics)
```
kanatohodets:
```suggestion
```

kanatohodets: Looks like leftover debug prints.

```suggestion
```

app/zipper/http_handlers.go
```diff
@@ -147,8 +147,10 @@ func findHandler(w http.ResponseWriter, req *http.Request) {
...
+	fmt.Println(err)
```
kanatohodets:
```suggestion
```

cmd/carbonzipper/main.go
```diff
@@ -67,7 +68,16 @@ func main() {
...
+		zap.String("build_version", BuildVersion),
+		zap.Any("zipperConfig", config),
+	)
 
-	zipper.StartCarbonZipper(config, logger)
+	app, err := zipper.InitializeApp(config, logger, BuildVersion)
+	if err != nil {
+		logger.Error("Error initializing app")
+	}
+	app.StartCarbonZipper()
```
kanatohodets: nit: any reason for `Start` for carbonapi vs `StartCarbonZipper` here?

cmd/carbonapi/main.go
```diff
@@ -38,5 +41,14 @@ func main() {
 		)
 	}
 	fh.Close()
-	capi.StartCarbonapi(api, logger)
+	expvar.NewString("BuildVersion").Set(BuildVersion)
+	logger.Info("starting carbonapi",
+		zap.String("build_version", BuildVersion),
+		zap.Any("apiConfig", api),
+	)
+	app, err := capi.InitializeApp(api, logger, BuildVersion)
```
kanatohodets: matter of taste: personally I'd probably call this just "New", "Initialize" to me sounds like it's taking an empty struct and populating the fields, whereas this builds a new struct and returns it.

app/zipper/api.go
```diff
@@ -68,18 +71,11 @@ func StartCarbonZipper(config cfg.Zipper,logger *zap.Logger){
...
 
-	expvar.NewString("BuildVersion").Set(BuildVersion)
-	logger = zapwriter.Logger("main")
-	logger.Info("starting carbonzipper",
-		zap.String("build_version", BuildVersion),
-		zap.Any("config", config),
-	)
-
-	runtime.GOMAXPROCS(config.MaxProcs)
+	runtime.GOMAXPROCS(app.config.MaxProcs)
```
kanatohodets: this kind of runtime-wide configuration is better left in a main.go setting, I think. It's not super nice behavior for a library to modify this kind of thing behind your back.

app/carbonapi/http_handlers_test.go
```diff
@@ -16,9 +16,8 @@ func TestShouldNotBlock(t *testing.T) {
 
 	req.Header.Add("foo", "bar")
 	rule := Rule{"foo": "block"}
-	config.blockHeaderRules = RuleConfig{Rules: []Rule{rule}}
 
-	if shouldBlockRequest(req) {
+	if shouldBlockRequest(req, []Rule{rule}) {
```
kanatohodets: This looks like a pre-existing condition, but I think either the func name or the error message are wrong: "shouldBlockRequest" returns true, we error saying "should not have blocked request"?

app/carbonapi/http_handlers.go
```diff
@@ -993,11 +993,11 @@ func blockHeaders(w http.ResponseWriter, r *http.Request) {
 	if len(m) == 0 {
 		logger.Error("couldn't create a rule from params")
 	} else {
-		fileData, err := loadBlockRuleConfig()
+		fileData, err := loadBlockRuleConfig(app)
 		if err == nil {
 			yaml.Unmarshal(fileData, &ruleConfig)
 		}
-		err1 = appendRuleToConfig(ruleConfig, m, logger)
+		err1 = appendRuleToConfig(ruleConfig, m, logger, app)
```
kanatohodets: I think it'd be a bit better to just pass in the `app.config.BlockHeaderFile` here, since that appears to be the only field used under the `appendRuleToConfig` call tree.

app/carbonapi/api.go
```diff
@@ -245,42 +247,42 @@ func registerPromehteusMetrics(logger *zap.Logger) {
...
 			}
 		} else {
-			config.blockHeaderRules = RuleConfig{}
+			app.blockHeaderRules = RuleConfig{}
 		}
 	}
 }
 
-func loadBlockRuleConfig() ([]byte, error) {
+func loadBlockRuleConfig(app *App) ([]byte, error) {
```
kanatohodets: like elsewhere, I think makes sense to pass it only `app.config.BlockHeaderFile` here, to keep it clear that this function is fairly self-contained

app/carbonapi/api.go
```diff
@@ -245,42 +247,42 @@ func registerPromehteusMetrics(logger *zap.Logger) {
...
-		fileData, err := loadBlockRuleConfig()
+		fileData, err := loadBlockRuleConfig(app)
 
 		if err == nil {
 			err = yaml.Unmarshal(fileData, &ruleConfig)
 			if err != nil {
 				logger.Error("couldn't unmarshal block rule file data")
 			} else {
-				config.blockHeaderRules = ruleConfig
+				app.blockHeaderRules = ruleConfig
```
kanatohodets: I think if a func is mutating the `app` struct, it should probably be a method hanging off that struct, rather than a function which takes the struct as an argument.

Also, I'm not sure about the context here, but now that our HTTP handlers are calling methods on the struct, we need to be really paranoid about any code which mutates fields on the app struct, since they can represent a data race.

app/carbonapi/api.go
```diff
@@ -218,21 +220,21 @@ func StartCarbonapi(api cfg.API, logger *zap.Logger) {
 	}
 }
 
-func registerPromehteusMetrics(logger *zap.Logger) {
+func registerPromehteusMetrics(logger *zap.Logger, app *App) {
```
kanatohodets: would it make sense for this to be a method on the app? it looks like it uses a fair number of fields, especially `initHandlersInternal`.

cmd/carbonzipper/main.go
gksinghjsr: no, there isn't. Will update them to be the same.

PR #14, opened 2018-10-29
fix empty result in moving functions

msaf1980:
On empty set panic 'index out of range' at expr/functions/moving/function.go:72


gksinghjsr:
@msaf1980 any particular reason why you closed the pull request? Looks like this could be useful for us as well.

PR #15, opened 2018-11-06
fix movingXyz error on intervals greater than 30 days

msaf1980:
Becouse data was sampled to 10 min on query with long data range (over 30 days), for functions movingXyz (like movingMax(Series, '3min'))  windowSize == 0 and I'm get panic at (*Windowed).Push.

### DIFF
--- a/expr/functions/moving/function.go
+++ b/expr/functions/moving/function.go
@@ -86,30 +86,38 @@ func (f *moving) Do(e parser.Expr, from, until int32, values map[parser.MetricRe
 		r.StopTime = until
 
 		for i, v := range a.Values {
-			if a.IsAbsent[i] {
-				// make sure missing values are ignored
-				v = math.NaN()
-			}
+			if windowSize == 0 {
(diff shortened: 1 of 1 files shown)


gunnihinn:
Thanks for the patch.

Do you think you could give me either an example of a query that causes this failure, or a test case that exhibits it? I'd like to get a regression test in place for this and am not sure that I'll correctly guess at the code path you hit in this problem.

msaf1980:
Hi, sorry for delay

For example 
movingMax(minSeries(*.cluster_health.status),'3min')

If we qeuery with long data range (over 30 days), returned points was sampled to 10 min interval

Some variables:
n = 180
int(arg[0].StepTime) = 600

windowSize = 180 / 600 = 0

So, empty slice was allocated at
w := &types.Windowed{Data: make([]float64, windowSize)}

And get a panic at Windowed.Push

With patch function returns result like graphite-web -  array of NaN values

PR #16, opened 2018-11-06
Fix moving empty

msaf1980:
On empty responce (no metrics) we get panic on moving and movingMedia functions. Add check for this.


gksinghjsr:
LGTM

PR #19, opened 2018-11-08
Add a minValue option to nonNegativeDerivative and perSecond

mbarbon:
It works in a way similar to maxValue: when the counter wraps, instead of
producing a null value, it computes the difference assuming the counter wrapped
to minValue.

This implements the same logic as was added to Graphite in
PR #2375 (https://github.com/graphite-project/graphite-web/pull/2375).


azhiltsov:
We need to update the functions api description for the function with this change.

PR #24, opened 2018-12-13
Close long-living idle connections

gunnihinn:
Since rolling out the pkg/backend refactoring to carbonzipper, we've seen a 100x increase in the number of open TCP connections per frontend host. This is accompanied by a similar increase in running goroutines, most of which are waiting on read- and writeLoops in the net/http library code.

I'm not able to explain this increase by staring at the code, and it seems to cause some problems for our surviving Python frontends when they try to connect to carbonzipper.

These couple of patches attempt to address the issue in two ways: We are extra careful about reading and closing response bodies, as not doing that is one source of TCP connection leaking in net/http; and we impose a timeout of 3 seconds on idle connections, after which they will be closed.

In testing, this has shown to be sufficient to get rid of the symptoms of increased connections and goroutines, but we remain ignorant of their root cause.

### DIFF
--- a/pkg/backend/net/net.go
+++ b/pkg/backend/net/net.go
@@ -194,19 +194,22 @@ func (b Backend) do(ctx context.Context, trace types.Trace, req *http.Request) (
 	t0 := time.Now()
 	resp, err := b.client.Do(req)
 	trace.AddHTTPCall(t0)
+
+	var body []byte
+	var bodyErr error
+	if resp != nil && resp.Body != nil {
+		t1 := time.Now()
+		body, bodyErr = ioutil.ReadAll(resp.Body)
+		resp.Body.Close()
+		trace.AddReadBody(t1)
+	}
+
 	if err != nil {
-		if resp != nil && resp.Body != nil {
-			resp.Body.Close()
-		}
 		return "", nil, err
 	}
 
-	t1 := time.Now()
-	body, err := ioutil.ReadAll(resp.Body)
-	resp.Body.Close()
-	trace.AddReadBody(t1)
-	if err != nil {
-		return "", nil, err
+	if bodyErr != nil {
+		return "", nil, bodyErr
 	}
 
 	if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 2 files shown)


pkg/backend/net/net.go
```diff
@@ -194,21 +194,24 @@ func (b Backend) do(ctx context.Context, trace types.Trace, req *http.Request) (
...
+		body, bodyErr = ioutil.ReadAll(resp.Body)
+		resp.Body.Close()
+		trace.AddReadBody(t1)
 	}
 
-	t1 := time.Now()
-	body, err := ioutil.ReadAll(resp.Body)
-	resp.Body.Close()
-	trace.AddReadBody(t1)
 	if err != nil {
```
icanhazbroccoli: [minor] I would suggest moving this check as early as 10 lines up

gunnihinn: ... before the [`client.Do`](https://github.com/bookingcom/carbonapi/blob/67b6cce823718b6272c15fc1becc5247d39222ec/pkg/backend/net/net.go#L195) call? I think I misunderstood something here.

icanhazbroccoli: I meant: right after line 197: trace.AddHTTPCall(t0). [This is the revision](https://github.com/bookingcom/carbonapi/blob/67b6cce823718b6272c15fc1becc5247d39222ec/pkg/backend/net/net.go) I'm checking.

gunnihinn: I want to believe you are right, but I think I need to either go read some stdlib code or bang out some examples to figure out what's going on.

The `net/http` documentation does say:

> It is the caller's responsibility to close Body. The default HTTP client's Transport may not reuse HTTP/1.x "keep-alive" TCP connections if the Body is not read to completion and closed.

It also does **not** say whether when we write
```
resp, err := client.Do(req)
```
then `resp` and `err` can both be non-nil, or when what would happen. If that can happen, then we'd need to close `resp.Body` before returning from the error.

There are [some sources](http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/) out there who claim this can happen:

> Most of the time when your http request fails the resp variable will be nil and the err variable will be non-nil. However, when you get a redirection failure both variables will be non-nil. This means you can still end up with a leak.

The code I wrote is the most paranoid reaction to the `net/http` docs and the statements in that article. It **will** read all of `resp.Body` and close it if it's ever non-nil.

gksinghjsr: "On error, any Response can be ignored." - https://golang.org/pkg/net/http/#Client.Do
Also this [stackoverflow answer](https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body-in-golang) talks about this.

icanhazbroccoli: Oh, I was not aware of this case. Thanks a lot for clarifying. Your original code now makes a perfect sense to me.

ksurent: Ooh, that's a good catch! I can think of a few places where this should be checked in other codebases...

PR #32, opened 2019-01-15
Config fix and cleanup

### LINKED ISSUE #27: Fix config
## 1. Do .toml configs work?

When trying to load a `.toml` config, I get an error:

    2019-01-13T13:50:55.626+0100    FATAL   main    Failed to parse config file     {"error": "yaml: unmarshal errors:\n  line 5: cannot unmarshal !!str `concure...` into cfg.preAPI"}

Also, I see only `yaml` annotations everywhere in config code.

If `.toml` configs don't work, we should consider removing them.

## 2. Current config does not work out-of-the box

Probably needs to be tweaked.

grzkv:
Fixes #27 

* Removed old configs that did not work
* Added new working configs
* Removed .toml configs, these were not even parsed
* Added simple run script to run both `carbonapi` and `zipper`
* Small readme update

Also, see commit messages for description.


config/carbonapi.yaml
```diff
@@ -35,23 +37,16 @@ tz: ""
...
 #
 # For some backends (e.x. graphite-clickhouse) you might want to set it to some insanly high value, like 100000
-
-# If 'true', carbonapi will send requests as is, with globs and braces
-# even without checking maxBatchSize
-# in result it can increase performance for graphite-clickhouse by removing /metric/find request before calling /render
-alwaysSendGlobsAsIs: false
-
-functionsConfigs:
-    graphiteWeb: ./graphiteWeb.example.yaml
```
azhiltsov: We need to keep config parameters even if we don't use them, this increases visibility for those who can't read the code about available options. You can comment it out instead

config/carbonapi.yaml
```diff
@@ -64,48 +59,107 @@ graphite:
...
+#    concurrencyLimitPerServer: 1200
+    keepAliveInterval: "30s"
+    maxIdleConnsPerHost: 1030
+    backends:
+      - http://127.0.0.1:8000
+      #TODO change back to go -carbon settings
+#      - http://127.0.0.1:8080
+#      - http://127.0.0.1:8082
+#    backendsv2:
+#        maxBatchSize: 300
```
azhiltsov: I am not sure that we are currently supporting such format in config

config/carbonapi.yaml
```diff
@@ -64,48 +59,107 @@ graphite:
...
     # This is the *ONLY* config element in this section that MUST be specified.
-    backends:
-        - "http://127.0.0.2:8080"
-        - "http://127.0.0.3:8080"
-        - "http://127.0.0.4:8080"
+#    backends:
+#        - "http://127.0.0.1:8080"
+#        - "http://127.0.0.1:8082"
+
+#    carbonsearchv2:
```
azhiltsov: carbonsearch was removed as I recall

config/carbonzipper.conf
```diff
@@ -1,4 +1,5 @@
-listen: ":8080"
+listen: ":8000"
+listenInternal: ":7000"
```
azhiltsov: this require a comment about what does it do and for what purpose can be used

config/carbonapi.yaml
grzkv: I have added these back as comments.

Issue #37, opened 2019-01-17
Move prometheus metrics settings to config files


grzkv:
Prometheus metrics parameters are hard-coded. They can only be changed by changing the code. We need to move them to config.

We need to do small refactoring to make this happen.

PR #38, opened 2019-01-17
Added time a request spends in queue to Prometheus metrics + Config refactoring

grzkv:
See title.

The parameters are hard-coded for now, this is described in #37 and will be addressed shortly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@
 *.out
 TODO
 *.log
+coverage.txt
 
 .vscode
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ before_install:
 
 script: make all test vet
 
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
+
 matrix:
   include:
     - go: 1.10.x
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ lint:
 	gometalinter --vendor --deadline=150s --cyclo-over=15 --exclude="\bexported \w+ (\S*['.]*)([a-zA-Z'.*]*) should have comment or be unexported\b" ./...
 
 test:
-	$(PKGCONF) $(GO) test ./... -race -coverprofile cover.out
+	$(PKGCONF) $(GO) test ./... -race -coverprofile=coverage.txt -covermode=atomic
 
 clean:
 	rm -f carbonapi carbonzipper
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Build Status](https://travis-ci.com/bookingcom/carbonapi.svg?branch=master)](https://travis-ci.com/bookingcom/carbonapi)
 [![Go Report Card](https://goreportcard.com/badge/github.com/bookingcom/carbonapi)](https://goreportcard.com/report/github.com/bookingcom/carbonapi)
+[![codecov](https://codecov.io/gh/bookingcom/carbonapi/branch/master/graph/badge.svg)](https://codecov.io/gh/bookingcom/carbonapi)
 
 Carbonapi is a Go-based Graphite frontend. It provides two binaries,
 `carbonapi` and `carbonzipper`, that unify responses from multiple Graphite


grzkv:
This also resolves #40

grzkv:
Parameters moved to config for `carbonapi`. Will update `carbonzipper` in the next PR.

app/carbonapi/app.go
```diff
@@ -51,43 +52,62 @@ type App struct {
...
-		},
-	),
-	DurationsLin: prometheus.NewHistogram(
-		prometheus.HistogramOpts{
-			Name:    "http_request_duration_seconds_lin",
-			Help:    "The duration of HTTP requests (linear)",
-			Buckets: prometheus.LinearBuckets(0.0, (50 * time.Millisecond).Seconds(), 40), // Up to 2 seconds
-		},
-	),
+	TimeInQueue  prometheus.Histogram
```
svagner: Why we need Histogram? Why no SummaryVec?

app/carbonapi/app.go
```diff
@@ -51,43 +52,62 @@ type App struct {
...
+				Help:    "The duration of HTTP requests (linear)",
+				Buckets: prometheus.LinearBuckets(0.0, (50 * time.Millisecond).Seconds(), 40), // Up to 2 seconds
+			},
+		),
+		TimeInQueue: prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Name: "time_in_queue",
+				Help: "Time a request to backend spends in queue, in ms",
+				Buckets: prometheus.LinearBuckets(
+					config.Zipper.Common.Monitoring.TimeInQueueHistogram.Start,
```
svagner: May be there is better to use getter got get config values?

app/carbonapi/app.go
```diff
@@ -42,6 +42,7 @@ import (
 // BuildVersion is provided to be overridden at build time. Eg. go build -ldflags -X 'main.BuildVersion=...'
 var BuildVersion string
```
svagner: We have public BuildVersion in package carbonapi as well. Why we need duplicates?

app/carbonapi/app.go
```diff
@@ -169,17 +186,19 @@ var zipperMetrics = struct {
...
 
-func New(api cfg.API, logger *zap.Logger, buildVersion string) (*App, error) {
-	if len(api.Backends) == 0 {
+// New creates a new app
+func New(config cfg.API, logger *zap.Logger, buildVersion string) (*App, error) {
+	if len(config.Backends) == 0 {
 		logger.Fatal("no backends specified for upstreams!")
 	}
 
 	BuildVersion = buildVersion
```
svagner: Passing buildVersion to function for set global isn't good idea:)

app/carbonapi/app.go
```diff
@@ -529,7 +550,10 @@ func deferredAccessLogging(r *http.Request, accessLogDetails *carbonapipb.Access
 		accessLogger.Info("request served", zap.Any("data", *accessLogDetails))
 		apiMetrics.Responses.Add(1)
 	}
-	prometheusMetrics.Responses.WithLabelValues(fmt.Sprintf("%d", accessLogDetails.HttpCode), accessLogDetails.Handler).Inc()
+
+	if app != nil {
```
svagner: You can't call method for struct App, if App eq nil. I think this condition can be rid.

app/carbonapi/app.go
```diff
@@ -529,7 +550,10 @@ func deferredAccessLogging(r *http.Request, accessLogDetails *carbonapipb.Access
...
+		app.prometheusMetrics.Responses.WithLabelValues(fmt.Sprintf("%d", accessLogDetails.HttpCode), accessLogDetails.Handler).Inc()
```
svagner: Here is better way to use strconv.Itoa. See https://gist.github.com/evalphobia/caee1602969a640a4530

app/carbonapi/app.go
```diff
@@ -581,8 +605,8 @@ func (app *App) bucketRequestTimes(req *http.Request, t time.Duration) {
 	expBucketIdx := findBucketIndex(expTimeBuckets, expBucket)
 	atomic.AddInt64(&expTimeBuckets[expBucketIdx], 1)
 
-	prometheusMetrics.DurationsExp.Observe(t.Seconds())
-	prometheusMetrics.DurationsLin.Observe(t.Seconds())
+	app.prometheusMetrics.DurationsExp.Observe(t.Seconds())
```
svagner: I think ir is better to measure duration in Milliseconds.

app/carbonapi/http_handlers.go
```diff
@@ -350,6 +350,9 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request) {
 					bs := backend.Filter(app.backends, request.Targets)
 					metrics, err := backend.Renders(ctx, bs, request)
 
+					// time in queue is converted to ms
+					app.prometheusMetrics.TimeInQueue.Observe(float64(request.Trace.Report()[2]) / 1000)
```
svagner: It would be great to make return value from Report method as some struct (not array) and make some getters for structure fields. It would be better to make source more readable.

app/carbonzipper/app.go
```diff
@@ -277,8 +283,8 @@ func (app *App) bucketRequestTimes(req *http.Request, t time.Duration) {
 	expBucketIdx := findBucketIndex(expTimeBuckets, expBucket)
 	atomic.AddInt64(&expTimeBuckets[expBucketIdx], 1)
 
-	prometheusMetrics.DurationsExp.Observe(t.Seconds())
-	prometheusMetrics.DurationsLin.Observe(t.Seconds())
+	app.prometheusMetrics.DurationsExp.Observe(t.Seconds())
```
svagner: Same thing as in comment about Seconds/Milliseconds

app/carbonzipper/http_handlers.go
```diff
@@ -267,7 +186,7 @@ func (app *App) renderHandler(w http.ResponseWriter, req *http.Request) {
 			zap.Error(err),
 		)
 		Metrics.Errors.Add(1)
-		prometheusMetrics.Responses.WithLabelValues(fmt.Sprintf("%d", http.StatusBadRequest), "render").Inc()
+		app.prometheusMetrics.Responses.WithLabelValues(fmt.Sprintf("%d", http.StatusBadRequest), "render").Inc()
```
svagner: Sprintf/Itoa

app/carbonapi/app.go
grzkv: Because we want to have more fine-graining and want to use Prometheus primitives.

app/carbonapi/app.go
grzkv: Yes, or event `viper`, but this is a big change and I am not sure it's worth it.

app/carbonapi/app.go
grzkv: Good point. We will look into it in the next refactoring step.

app/carbonapi/app.go
grzkv: I can, and it actually happened during tests. I have added this line after a test failed because of `app == nil`.

grzkv:
@svagner Here's the situation: the code has a lot of problems and I am addressing the biggest ones first. At the same time, we need to make incremental changes, so I cannot make too many changes at once.

`gometalinter` has around 200 warnings now.

It will take probably months until the code is of acceptable quality.

app/carbonapi/app.go
```diff
@@ -51,43 +52,62 @@ type App struct {
 	defaultTimeZone *time.Location
 
 	backends []backend.Backend
+
+	prometheusMetrics PrometheusMetrics
```
icanhazbroccoli: It's a very generic suggestion, but I believe it still has a place here: I might suggest isolating Prometheus in a stand-alone module and implementing an adapter layer. The motivation for this change is to avoid the technology lock in the future.

app/carbonapi/app.go
```diff
@@ -51,43 +52,62 @@ type App struct {
...
-		prometheus.HistogramOpts{
-			Name:    "http_request_duration_seconds_lin",
-			Help:    "The duration of HTTP requests (linear)",
-			Buckets: prometheus.LinearBuckets(0.0, (50 * time.Millisecond).Seconds(), 40), // Up to 2 seconds
-		},
-	),
+	TimeInQueue  prometheus.Histogram
+}
+
+func newPrometheusMetrics(config cfg.API) PrometheusMetrics {
```
icanhazbroccoli: I'm wondering if this is a private constructor and is supposed to be used from the inside of the module?

app/carbonapi/http_handlers.go
```diff
@@ -617,7 +620,7 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request) {
 
 	logAsError := false
 	defer func() {
-		deferredAccessLogging(r, &accessLogDetails, t0, logAsError)
+		app.deferredAccessLogging(r, &accessLogDetails, t0, logAsError)
```
icanhazbroccoli: It can always be called simply as:
defer app.deferredAccessLogging(...), or, as I suggested earlier: defer deferredAccessLogging(app, ...)

app/carbonapi/http_handlers.go
```diff
@@ -658,7 +661,7 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request) {
 
 			http.Error(w, msg, code)
 			apiMetrics.Errors.Add(1)
-			prometheusMetrics.Responses.WithLabelValues(fmt.Sprintf("%d", code), "find").Inc()
+			app.prometheusMetrics.Responses.WithLabelValues(fmt.Sprintf("%d", code), "find").Inc()
```
icanhazbroccoli: Consider using strconv.Itoa(int) instead. This module provides some conversion semantics, not string formatting.

app/carbonapi/http_handlers.go
```diff
@@ -802,7 +805,7 @@ func (app *App) infoHandler(w http.ResponseWriter, r *http.Request) {
 
 	logAsError := false
 	defer func() {
-		deferredAccessLogging(r, &accessLogDetails, t0, logAsError)
+		app.deferredAccessLogging(r, &accessLogDetails, t0, logAsError)
```
icanhazbroccoli: What took my attention is: this defer call is spread all around the code, even though it's always being called with the same set of arguments. I'm curious if this can be put as a single statement in the very end of the method call and, maybe, this can un-defer it entirely.

app/carbonzipper/http_handlers.go
```diff
@@ -289,7 +208,7 @@ func (app *App) renderHandler(w http.ResponseWriter, req *http.Request) {
...
```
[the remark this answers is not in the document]
icanhazbroccoli: Another suggestion: as code is always an HTTP code (this assumption might be terribly wrong and is only based on what I've observed in this PR), consider refactoring app.prometheusMetrics.Responses.WithLabelValues argument types to: (int, string)

app/carbonapi/app.go
grzkv: Thanks for the suggestion. We will consider doing this.

app/carbonapi/app.go
grzkv: It is. And it's not exported. What is your point then?

app/carbonapi/http_handlers.go
grzkv: Yes, you are 100% right. This needs to be cleaned up altogether though. This will be taken care of in one of the following refactorings.

Thanks for pointing this out.

app/carbonapi/http_handlers.go
grzkv: +1 will do in one of the next refactorings

app/carbonapi/http_handlers.go
grzkv: You are probably right. This code needs a big refactoring, and this will be part of it.

app/carbonzipper/http_handlers.go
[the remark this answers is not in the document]
grzkv: Makes sense to me. Thanks.

PR #39, opened 2019-01-17
Add carbonapi_uuid to http response header

### LINKED ISSUE #36: Expose carbonapi_uuid in http response headers
Presenting carbonapi_uuid to a user with result of render will allow easier troubleshooting by user.
Suggested name is X-Carbonapi-uuid

azhiltsov:
Fixes #36

### DIFF
--- a/app/carbonapi/http_handlers.go
+++ b/app/carbonapi/http_handlers.go
@@ -127,8 +127,9 @@ func initHandlers(app *App) http.Handler {
 	return r
 }
 
-func writeResponse(w http.ResponseWriter, b []byte, format string, jsonp string) {
+func writeResponse(w http.ResponseWriter, b []byte, format string, jsonp string, ctx context.Context) {
 
+	w.Header().Set("X-Carbonapi-UUID", util.GetUUID(ctx))
 	switch format {
 	case jsonFormat:
 		if jsonp != "" {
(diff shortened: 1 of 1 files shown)


app/carbonapi/http_handlers.go
```diff
@@ -127,8 +127,9 @@ func initHandlers(app *App) http.Handler {
 	return r
 }
 
-func writeResponse(w http.ResponseWriter, b []byte, format string, jsonp string) {
+func writeResponse(w http.ResponseWriter, b []byte, format string, jsonp string, ctx context.Context) {
 
+	w.Header().Set("X-carbonapi-uuid", util.GetUUID(ctx))
```
grzkv: I think, "X-Carbonapi-UUID" would be a more canonical name.

grzkv:
It can be that the exact placing is not perfect, but this part of the code needs refactoring anyway. So, I think it's perfectly fine for now.

azhiltsov:
@grzkv The way I passed a context, is it fine?
Do you have a suggestion how to improve it?

app/carbonapi/http_handlers.go
azhiltsov: I agree

grzkv:
> @grzkv The way I passed a context, is it fine?
> Do you have a suggestion how to improve it?

Just as I said in the previous comment. It looks perfectly fine to me. We will make the structure better during refactoring.

PR #42, opened 2019-01-19
Mimicking graphite-web margin indentation for graphOnly=1

### LINKED ISSUE #41: carbonapi png mismatches graphite-web png ig graphOnly=1
When you render a graphOnly=1 png there is a space on the left in carbonapi, but no space in graphite-web.
Also this make impossible to render smaller graphs i.e. width=50 height=20

azhiltsov:
Fixes #41
https://github.com/graphite-project/graphite-web/blob/0.9.12/webapp/graphite/render/glyph.py#L560-L563


grzkv:
Are you sure this fixes it? When we looked through the code last time, the parameter was in more than one place.

Did you test this through?

grzkv:
IMO, even if it does not completely fix the problem, chances are that it makes us closer to the solution. Makes sense to add this, but test thoroughly after rollout.

Issue #58, opened 2019-01-28
Dockerize carbonapi setup


grzkv:
Make a Docker config that would include the following:

1. `carbonapi`
2. `carbonzipper`
3. `go-carbon`
4. Some data pushed to storage to play with

This will:
- allow other people to try `carbonapi` easily, and contribute as well
- will make system testing possible, inside CI as well
- will make manual testing easier

PR #85, opened 2019-02-14
Added docker configs for all essential services + docker compose setup for the full stack

grzkv:
see title + commit messages

Now the full stack of 
- `go-carbon`
- `carbonzipper`
- `carbonapi`
can be run with single `docker-compose`


grzkv:
Fixes #58

README.md
```diff
@@ -34,9 +34,25 @@ make debug
 We do not provide packages for install at this time. Contact us if you're
 interested in those.
 
-### Run
+## Run
 
-You can easily run the program with sample config provided in `config/` directory by running `./run.sh`.
+Run the full stack `carbonapi` -> `zipper` -> `go-carbon` with:
```
azhiltsov: in this setup you don't need a zipper.
carbonapi -> go-carbon should work fine. Zipper needed if you have more than 1 go-carbon instance.

gksinghjsr: No, we should have our setup as the default, and not the simplest stack missing a key component.

grzkv: It makes sense to have a simplified setup as a special case.

But in general case, we need `zipper` to have a full stack for testing changes and running system tests.

PR #94, opened 2019-03-01
Add logging when render request contains metrics with different resolution

gksinghjsr:


grzkv:
Thanks, @gksinghjsr for this. I think, this a much-needed check!

My only concern is performance. The number of ops we do to do the checks can be quite high.

We need to add benchmarks for this.

grzkv:
#95

PR #100, opened 2019-03-18
Script to generate and populate short lived, high cardinality pod metrics.

gksinghjsr:


tests/short_lived_metrics_gen.py
```diff
@@ -0,0 +1,93 @@
...
+
+parser=argparse.ArgumentParser()
+
+parser.add_argument('--host', default='localhost')
+parser.add_argument('--port', default=2003, type=int)
+parser.add_argument('--datapoints', default=1440*3, type=int)
+parser.add_argument('--paths-outer', default=100, type=int)
+parser.add_argument('--paths-inner', default=100, type=int)
+parser.add_argument('--total-metrics', default=100, type=int)
+parser.add_argument('--total-pods', default=100, type=int)
```
grzkv: I would rename this to make it more generic. Same about other variations. Maybe something like *first*, *second*. *third*, *fourth*? Different people may want to use this differently.

tests/short_lived_metrics_gen.py
```diff
@@ -0,0 +1,93 @@
...
+parser.add_argument('--outfile', default="system_test.csv")
+parser.add_argument('--stream', default=1, type=int)
+parser.add_argument('--batches', default=2, type=int, help='total batches of metrics')
```
grzkv: I don't quite understand why'd we want this?

PR #113, opened 2019-04-05
Errors refactoring

grzkv:
- extracted error logic from shared code to `api` and `zipper` separately
- added unit tests for errors fan-in
- changed error logic for `finds`
    - all not-found -> not found
    - all mixed errors -> internal server error
    - some backends succeeded -> success
    - not found returns `200` code

See commit messages for more info

### DIFF
--- a/app/carbonzipper/http_handlers.go
+++ b/app/carbonzipper/http_handlers.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/bookingcom/carbonapi/pkg/backend"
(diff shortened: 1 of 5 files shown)


app/carbonzipper/http_handlers.go
```diff
@@ -34,12 +35,65 @@ const (
...
+				break
+			}
+		}
+
+		if allErrorsNotFound {
+			return types.ErrNotFound("all backends returned not found")
+		}
+
+		// if it's not a not-found, it's an internal error
+		logger.Error("Requests from all backends failed",
```
gksinghjsr: Shouldn't it log the this first, before the return above.

grzkv: No, because we only want the log when we return generic error. This way we have no error log when it's a not found, which is good. This way we get much cleaner logs.

PR #125, opened 2019-04-15
Next stage of (error) refactoring

grzkv:
- not-found during `/find` request is not considered an error anymore both on `api` and `zipper`
- added metric for not founds on `api`
- fixed double-counting of requests on `zipper` `/find` in case of not-found
- removed obsolete fan-ins in `api` handlers
- broke `/render` handler in *api* into smaller pieces


app/carbonapi/http_handlers.go
```diff
@@ -428,37 +359,143 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request) {
...
+	switch form.format {
 	case jsonFormat:
 		if maxDataPoints, _ := strconv.Atoi(r.FormValue("maxDataPoints")); maxDataPoints != 0 {
 			types.ConsolidateJSON(maxDataPoints, results)
 		}
 
 		body = types.MarshalJSON(results)
 	case protobufFormat, protobuf3Format:
-		body, err = types.MarshalProtobuf(results)
+		body, err := types.MarshalProtobuf(results)
```
gksinghjsr: Won't this shadow the variable created in outer scope and create issues?

grzkv: This will shadow, but I don't see problem with that.

gksinghjsr: https://play.golang.org/p/qkUigxU9eFN

grzkv: Yes, but what's the problem? Shadowing value is only used inside single case as expected.

gksinghjsr: body would return nil in case format is one of protobufFormat, protobuf3Format

grzkv: If you talk about body, then yes, it is a problem.

Issue #110, opened 2019-04-01
Zipper tries to send metrics to graphite even when it's not specified in config


grzkv:
When there is no `graphite` option specified in `zipper` config, it tries to push metrics anyway to *127.0.0.1:3002*. If there is no graphite running there (and it's not supposed to be when there is nothing in the config), it logs packs of errors.

`Zipper` must not push to graphite when it is not specified in config.

grzkv:
Here's an example of errors logged:

```
2019/04/01 18:01:32 g2g: carbon.zipper.star.exp.requests_in_01600ms_to_03200ms: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.exp.requests_in_03200ms_to_06400ms: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.requests_in_800ms_to_900ms: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.uptime: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.total_alloc: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.info_errors: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.requests_in_300ms_to_400ms: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.exp.requests_in_00800ms_to_01600ms: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.exp.requests_in_25600ms_to_51200ms: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
2019/04/01 18:01:32 g2g: carbon.zipper.star.exp.requests_in_12800ms_to_25600ms: failed; reconnect attempt: dial tcp 127.0.0.1:2003: connect: connection refused
[...]

PR #126, opened 2019-04-15
No graphite proxy by default + no block rules file is empty set of rules

grzkv:
see commit messages for detail


grzkv:
Fixes #110

PR #134, opened 2019-04-24
Fix builds without cairo

reyjrar:
I'm not a golang developer, but after a lot of fighting with the build,
I *think* this restores the option to build this without libcairo.

### DIFF
--- a/expr/functions/cairo/png/pixel_ratio.go
+++ b/expr/functions/cairo/png/pixel_ratio.go
@@ -1,4 +1,5 @@
 // +build cairo
+
 package png
 
 import "github.com/evmar/gocairo/cairo"
--- a/expr/functions/cairo/png/pixel_ratio.go
+++ b/expr/functions/cairo/png/pixel_ratio.go
@@ -1,3 +1,4 @@
+// +build cairo
 package png
 
 import "github.com/evmar/gocairo/cairo"


grzkv:
@reyjrar The CI has to succeed before we have a look at this PR.

blind-oracle:
@reyjrar You need to add a newline after the **// +build ** for it to work.

grzkv:
On my machine the `nocairo` build works even without these changes.

On the first glance, it makes sense to exclude the compilation of the `pixel_ratio` when *cairo* tag is absent. But there may be deeper implications.

blind-oracle:
@grzkv *nocairo* build works if you have cairo lib present and links it anyway because `pixel_ratio.go` imports `github.com/evmar/gocairo/cairo` which is cgo.

I don't think that there are any implication in not building `pixel_ratio.go` if nothing except cairo-related stuff depends on it. And if there were anything then carbonapi would not have built without it.

reyjrar:
I'm building using mock, so it's a pristine environment every run and libcairo and friends are not around.  I'm building this for CentOS 6, where the cairo version doesn't meet the minimum cairo version required.

koorgoo:
Checked the PR on CentOS 6 - it worked.

PR #135, opened 2019-04-25
Added sorting when returning json response in api

grzkv:
see title


grzkv:
@azhiltsov Sorry for the late response. The results of find were returned unsorted. This means, when browsing web UI or in Grafana, one could get a list of ~100 metrics to select from in unsorted order.

PR #140, opened 2019-05-16
Refactored render handler for api. The requests now fail per-target.

grzkv:
…e.g. if there are 3 targets and one fails, we still return a success with the data for 2 successful targets. Error handling is changed accordingly. Separate targets fail if at least one request fails with an error different from not-found.

### DIFF
--- a/app/carbonapi/http_handlers.go
+++ b/app/carbonapi/http_handlers.go
@@ -187,130 +187,21 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request) {
 		apiMetrics.RequestCacheMisses.Add(1)
 	}
 
-	var results []*types.MetricData
 	metricMap := make(map[parser.MetricRequest][]*types.MetricData)
 
-	var metrics []string
+	var results []*types.MetricData
 	var targetErrs []error
-	// TODO(gmagnusson): Put the body of this loop in a select { } and cancel work
+	// TODO (grzkv) Modification of *form* inside the loop is never applied
 	for _, target := range form.targets {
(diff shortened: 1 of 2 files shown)


app/carbonapi/http_handlers_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"net/http"
 	"testing"
 
-	"github.com/bookingcom/carbonapi/pkg/types"
+	typ "github.com/bookingcom/carbonapi/pkg/types"
```
koorgoo: Should we change a reference not within this commit?
I see no conflicts in changes. Let's keep the diff as small as possible.

grzkv: This is an ongoing change. In many places, little by little. This is just one step of many. So, let's keep it.

app/carbonapi/http_handlers.go
```diff
@@ -521,7 +526,7 @@ func (app *App) renderWriteBody(results []*types.MetricData, form renderForm, r
 	return body, nil
 }
 
-func errsFanIn(errs []error, n int) error {
+func optimistFanIn(errs []error, n int) error {
```
koorgoo: I suggest to change the func's signature to:
```
func expectNotFound(n int, errs []error) error {
  if len(errs) < n {
    return nil
  } 
  // continue without switch
}
```

app/carbonapi/http_handlers_test.go
koorgoo: Ok then.

PR #148, opened 2019-05-31
Gysingh/zipper tld

gksinghjsr:
Sets up probes to cache tld for each backend store in zipper. This enables cutting down unnecessary requests.


grzkv:
I'm sorry but I'd like to discuss this in a bit more detail.

azhiltsov:
Considering that carbonapi talks to zipper, what would be the zipper application state while it querying TLD? Would it listen the data port end reply to queries?

PR #150, opened 2019-06-03
Changed error merge logic in case of all errors from backends to zipper: now we return 500 iff majority of backends return 500

grzkv:
see commits


grzkv:
@gksinghjsr @azhiltsov Please, have a look.

PR #160, opened 2019-06-04
limit max time duration and response size for upstream query

msaf1980:
Limits for upstream query:
maxDuration - limit from/until time duration (for protect  from malformed requests with incorrect timestamps)
maxSize - limit for upstream response size

### DIFF
--- a/app/carbonapi/app.go
+++ b/app/carbonapi/app.go
@@ -538,6 +538,7 @@ func initBackend(config cfg.API, logger *zap.Logger) (backend.Backend, error) {
 		Limit:              config.ConcurrencyLimitPerServer,
 		PathCacheExpirySec: uint32(config.ExpireDelaySec),
 		Logger:             logger,
+		MaxSize:            config.Limits.MaxSize,
 	})
 
 	if err != nil {
--- a/app/carbonapi/app_test.go
+++ b/app/carbonapi/app_test.go
@@ -177,6 +177,20 @@ func setUpRequest(t *testing.T, url string) *http.Request {
 	return req
 }
 
+func TestInitBackends(t *testing.T) {
+	config := cfg.DefaultAPIConfig()
+	c := cfg.GetDefaultLoggerConfig()
+	c.Level = "none"
+	zapwriter.ApplyConfig([]zapwriter.Config{c})
+	logger := zapwriter.Logger("main")
+	config.Zipper.Common.Backends = append(config.Zipper.Common.Backends, "http://127.0.0.1:9090")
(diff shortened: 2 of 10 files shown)


grzkv:
Hi @msaf1980 

Thanks for the pull request. It consists of two parts, so I will discuss them separately.

### I. Limit max time window that can be queried

- the check you perform checks for maximum allowed timeframe, AFAIK
- putting a max cap on the amount of data returned is the job done by `go-carbon` and *whisper*. When querying big timeframe, resolution will be low and the amount of data will not overload the system. If you experience the problem of pumping through too much data, maybe you could address it be changing the settings of the storages;
- in general, we don't put a limit on the timeframe. I see nothing wrong with querying wide timeframe;
- I am not sure if the community would benefit from imposing a constraint by the time window size.

### II. Limit max returned data

It is not clear to me what is the benefit of this change since you check for the amount of data when it was already acquired, so the majority of the effort was already done.

msaf1980:
I. Limit max time window that can be queried
Pre-history of this - some people in our enviriment insert wrong timestamp (lost last number in from timestamp). Result - date in 1970 year.

II. Limit max returned data
This is additional protection, also useful check limits on storage level - may reduce I/O and CPU usage on storage

Additional reason for this patches - human readable errors in grafana dashboards (insead of unnamed Request Error - now carbonapi don't read error text from storage backend)

grzkv:
> Pre-history of this - some people in our environment insert wrong timestamp (lost the last number in from timestamp). The result - date in the 1970 year.

This should be fixed with input validation rather than by restricting the time window in my opinion. See #171

> II. Limit max returned data. This is additional protection, more useful check limits on storage level - may reduce I/O and CPU usage on storage

See my earlier answer. This is applied after the storage already returned the data. It is not clear to me how this could reduce the load on storages.

msaf1980:
> > Pre-history of this - some people in our environment insert wrong timestamp (lost the last number in from timestamp). The result - date in the 1970 year.
> 
> This should be fixed with input validation rather than by restricting the time window in my opinion. See #171

May be. But restricting the time window more simple.
> 
> > II. Limit max returned data. This is additional protection, more useful check limits on storage level - may reduce I/O and CPU usage on storage
> 
> See my earlier answer. This is applied after the storage already returned the data. It is not clear to me how this could reduce the load on storages.

Check limits on storage level  may reduce I/O and CPU usage on storage.
Check on carbonapi is a additional protection (restrict time range for graphite-clickhouse is buggy - on default table time range not checked - possible, workaround exist in latest version - i'll see commit with simular functionality, but not tested it).

PR #162, opened 2019-06-05
updated documentation build steps with pkg-config note

KozzyKoder:
Build process requires pkg-config to be installed. Since it is not obvious step makes sense to include its installation steps into documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,20 @@ To build the binaries with debug symbols, run:
 make debug
 ```
 
+Note: build process requires pkg-config to be installed:
+
+### Mac OS X
+
+```
+curl https://pkgconfig.freedesktop.org/releases/pkg-config-0.29.tar.gz -o pkgconfig.tgz
+tar -zxf pkgconfig.tgz && cd pkg-config-0.29
+```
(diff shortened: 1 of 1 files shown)


grzkv:
@gksinghjsr Did you encounter any of the problems described in this addition?

KozzyKoder:
Yep, failed to install 2 times:
first time it failed because of glib dependency. Key --with-internal-glib solved this issue for me.
Next time it failed with this flag, since this flag requires env variable LDFLAGS to be set up.

README.md
```diff
@@ -31,6 +31,20 @@ To build the binaries with debug symbols, run:
 make debug
 ```
 
+Note: build process requires pkg-config to be installed:
+
+### Mac OS X
+
+```
+curl https://pkgconfig.freedesktop.org/releases/pkg-config-0.29.tar.gz -o pkgconfig.tgz
```
grzkv: These version numbers will not last.

grzkv:
How about removing specific version numbers? These will become obsolete soon.

grzkv:
I guess this should be described as a special case. I did not encounter problems like this when doing that on MacOS. How about you @gksinghjsr ?

README.md
KozzyKoder: removed specific version of pkg-config. User can specify version that fits him the best.

grzkv:
LGTM! Thanks for the pull-request!

PR #167, opened 2019-06-12
Added panic recoveries and logging in api

### LINKED ISSUE #163: Panic recovery on api
All panics on `api` have to be recovered from and logged. Specifically, we probably need to do that separately for handlers and globally.

grzkv:
- Added panic recovery per HTTP handler
- Added global panic logging and recovery in api

Fixes #163


grzkv:
@gksinghjsr Could you please have a look as well. Just to be safe. Thx.

PR #169, opened 2019-06-12
Gysingh/parse metrics

gksinghjsr:
Fixes parser wherein it was incorrectly parsing function parameters as metric names. This resulted in many unnecessary requests to backend stores.


grzkv:
It is a rather specific change. I guess a comment would be in order. 

Also, it seems like it should be pretty easy to add unit test to cover this case. What do you think?

PR #170, opened 2019-06-13
fixing errors in default config to reflect the state of code

azhiltsov:
removing unused idleConnections
placing keepAliveInterval from the upstreams section where it was
silently ignored the top


grzkv:
Have you tested this with docker-compose?

PR #173, opened 2019-06-22
Tests for request limitation by header functionality

KozzyKoder:
Extracted methods related to requests limitation by headers to separate package;
Separated config file operations and logic of limitation;
Added tests for package;


limiter/config_file.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+//iConfigFile interface provides abstraction for managing
```
grzkv: Why does this have *i* in front? Can it be for the type? See https://dave.cheney.net/2019/01/29/you-shouldnt-name-your-variables-after-their-types-for-the-same-reason-you-wouldnt-name-your-pets-dog-or-cat

limiter/request_limiter.go
```diff
@@ -0,0 +1,146 @@
+package limiter
```
grzkv: By convention, in this project, we call the request rate limiter a "limiter". Please, rename the package to something else, e.g. "blocker" or something.

limiter/request_limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	config       iConfigFile
+	logger       *zap.Logger
+	rules        atomic.Value
+	updatePeriod time.Duration
+}
+
+//NewRequestLimiter creates a new instance of request limiter without any rules
+//and sets name of config files that will be used as storage for rules
+func NewRequestLimiter(blockHeaderFile string, updatePeriod time.Duration, logger *zap.Logger) *RequestLimiter {
+	instance := &RequestLimiter{
```
grzkv: Does adding a separate object for limiter add any value? This does not solve the global vars problem because an instance is saved in App anyway.

limiter/config_file.go
KozzyKoder: Renamed to more natural name for interface

limiter/request_limiter.go
KozzyKoder: Renamed package to blocker

limiter/request_limiter.go
KozzyKoder: It does since fields inside object enclosed by package and not accessible outside of blocker package due to naming. Main purpose of this object is to manage access to rule values. Also was possible to define variables on level of package. But we need to initialise some of them before using this package(config file name, update period). Introduction of some sort Initialisation function could solve that and instead introduced implicit time restriction between call of Initialisation function and functions that manages rules. I followed explicit path there.

KozzyKoder:
this PR also fixes https://github.com/bookingcom/carbonapi/issues/197

grzkv:
@gksinghjsr Could you please have a quick look too?

PR #180, opened 2019-07-25
Graphite tags

msaf1980:
Initial support for graphite tags
At now function groupByTags support aggregation avg (average alias for compability with graphite-web), sum, min, max.

### DIFF
--- a/app/carbonapi/http_handlers.go
+++ b/app/carbonapi/http_handlers.go
@@ -1275,6 +1275,7 @@ supported requests:
 	/info/?target=
 	/functions/
 	/tags/autoComplete/tags
+	/tags/autoComplete/values
 `)
 
 func (app *App) usageHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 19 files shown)


msaf1980:
At now function groupByTags support aggregation avg (average alias for compability with graphite-web), sum, min, max.
Not realized median, diff, stddev, range, multiply. If needed, I can realize it.
https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.aggregate

azhiltsov:
@msaf1980 thank you for your contribution!

In order to make tags work in grafana [this](https://github.com/bookingcom/carbonapi/blob/master/app/carbonapi/http_handlers.go#L1294 
) also need to be fixed.
Additionally to that I would like to have a way to disable tag (or at least make absence if them invisible to an end-user if storage doesn't have such capability).
Might be tagsHandler could keep returning empty strings in such cases?
@grzkv @gysinghb could you suggest how this can be implemented?

msaf1980:
>  Additionally to that I would like to have a way to disable tag (or at least make absence if them invisible to an end-user if storage doesn't have such capability).
tagsEnable bool parameter in config ?

msaf1980:
Additional - for support tags functions in graphana query editor carbonapi need return version 'Graphite 1.1'

 tagHandler like this
https://github.com/go-graphite/carbonapi/blob/master/cmd/carbonapi/http/tags_handler.go
I'll port them

msaf1980:
Implement tagsHandler. Queries /tags/autoComplete/tags and /tags/autoComplete/values work, but in Grafana query editor seriesByTag not avaliable (got 'error parsing target: Method not found seriesByTag' in grafana front). functionHandler return seriesByTag.

PR #183, opened 2019-08-28
Prometheus latency metrics for render requests

gksinghjsr:
Adds prometheus latency metrics for render requests, dividing them into two branches, simple and complex, based on the heaviness of the requests, measured in total metrics queried.

### DIFF
--- a/app/carbonapi/http_handlers.go
+++ b/app/carbonapi/http_handlers.go
@@ -198,7 +198,7 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request) {
 			app.prometheusMetrics.RenderDurationPerPointExp.Observe(time.Since(t0).Seconds() * 1000 / float64(size))
 		}
 		//2xx response code is treated as success
-		if toLog.HttpCode > 0 && toLog.HttpCode/100 == 2 {
+		if toLog.HttpCode/100 == 2 {
 			if toLog.TotalMetricCount < int64(app.config.MaxBatchSize) {
 				app.prometheusMetrics.RenderDurationExpSimple.Observe(time.Since(t0).Seconds())
 			} else {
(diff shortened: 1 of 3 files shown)


app/carbonapi/http_handlers.go
```diff
@@ -193,9 +193,18 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request) {
 	var targetErrs []error
 	size := 0
 	defer func() {
+		//TODO: cleanup RenderDurationPerPointExp
 		if size > 0 {
 			app.prometheusMetrics.RenderDurationPerPointExp.Observe(time.Since(t0).Seconds() * 1000 / float64(size))
 		}
+		//2xx response code is treated as success
+		if toLog.HttpCode > 0 && toLog.HttpCode/100 == 2 {
```
grzkv: I guess this can be replaced with just second part. The second part is false if the first one is false.

app/carbonapi/http_handlers.go
```diff
@@ -193,9 +193,18 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request) {
...
+			if toLog.TotalMetricCount < int64(app.config.MaxBatchSize) {
```
grzkv: The `MaxBatchSize` may be dropped in quite near future when we optimize to not pre-send finds. This distinction will then lose the point.

azhiltsov: This can go to comments as TODO

app/carbonapi/http_handlers.go
azhiltsov: I agree with @grzkv

gksinghjsr: yes, it was added as just a safety check.

app/carbonapi/http_handlers.go
gksinghjsr: @grzkv even then it would remain as a boundary for classifying simple/complex queries.

Issue #70, opened 2019-02-03
Add full-stack system tests


grzkv:
Add automatic system test that:

1. Spins-up the setup from #58 
2. Feeds some mock data on the storage
3. Retrieves data via the `go-carbon`→`zipper`→`carbonapi` chain
4. Verifies correctness

grzkv:
This can be crucial to test *clickhouse* support.

grzkv:
The tests should be supported with various back-ends. In the beginning, both `graphite-clickhouse` and `go-carbon` should be supported.

grzkv:
The `docker-compose` to spin-off the system is in place, but there is still no automation. The automated version must not include `clickhouse` config.

PR #190, opened 2019-09-09
Switch to go modules

### LINKED ISSUE #131: Migrate to go modules
Currently we use `dep` for dependency management. As it becomes deprecated in favor of `go modules`, we need to migrate as well.

paskal:
Should fix #131.
Build and tests are working, `system_test.sh` executed successfully: please point me to how to test this properly.


grzkv:
Are you sure the libs in use all stay the same version?

paskal:
All directly mentioned libs are the same version. The problem I see is the removal of the many vendored directories: let's include all possible build tags (and with -mod=vendor) in CI pipeline?

grzkv:
I'd keep vendor for now.

grzkv:
If we keep vendor we can have a much better vision about what has changed in the used libs. What do you think, @paskal ?

paskal:
Done, moduels introduced and vendors are intact.

azhiltsov:
vendor was introduces in order to have reproduceabe builds. We can remove them only in case if modules can provide us the same guarantees.

paskal:
@azhiltsov, go modules can vendor dependencies as well but @grzkv is against introducing any changes to vendor folder (and go modules do introduce changes) so I left go.mod and didn't touch existing vendoring.
As I see the problem is the following: no one can be sure that some change leaves the project in working state and it even builds properly with build tag `cairo`. The only way to know for sure for now is to run it somewhere in production and see what happens. I hope #70 will fix that situation.

paskal:
Modules introduced with minimal changes to vendored versions possible, please review. I haven't switched travis and all the code to using vendored versions, that would be a different issue to work on.

grzkv:
Did you use any info from `dep` when migrating?

paskal:
Yes, the **only** difference is `yaml.v2` 2.2.2 vs 2.2.1 original: it adds single new function and I didn't revert this bugfix version upgrade.

grzkv:
I will take some time to reflect on this change. It looks good on first glance, but there may be some non-obvious discrepancies. Sorry that this is taking a bit longer.

paskal:
@grzkv I've updated and re-checked everything once more and also added and tested vendoring for builds and tests. Everything works the same way it works now, libraries versions are the same.

paskal:
@grzkv let's finally merge it?

paskal:
@grzkv @azhiltsov let's merge this one?

grzkv:
@paskal I will test this one following day or two.

grzkv:
@paskal I will test this one tomorrow or day after.

paskal:
@grzkv I believe you can do this!

harudark:
I'm still tryig to figure out what's the best way to migrate to go modules with vendoring.
Because @grzkv wants to move forward this PR today, I share some of my thoughts here.


According to [this](https://blog.golang.org/migrating-to-go-modules),
> It is always good practice to run go mod tidy before committing a go.mod file to version control

I understand this PR is trying to introduce modules with minimal changes to vendored versions possible.
But if `go mod tidy` hasn't been run, I don't think go modules switching has been completed.

The problem is if we run `go mod tidy` right now, `gonum.org/v1/gonum` will upgrade to v0.6.2(released on 25 Dec 2019), which is a big change compared to the current version "20181001095203".

My suggestion is it's better to wait for #70 to fix and then move to go modules switching.

paskal:
I agree that it's better to proceed with the switch when the project will really need it and will have enough confidence to do so rather than now. I am not willing to continue updating this branch and closing this PR.

Issue #149, opened 2019-05-31
Have immediate context cancellations in all handlers


grzkv:
In some handlers (e.g. `render` handler in `zipper`), the handlers do not return immediately if context was cancelled. This can lead to goroutines piling up if inflow of requests cannot be handled, among other things.

jwkohnen:
I have climbed the call stack from [`app/carbonzipper.App.renderHandler()`](https://github.com/bookingcom/carbonapi/blob/b5dd1535f91fd0e864c5bebf0e652ba0a98c8dfd/app/carbonzipper/http_handlers.go#L276) and came along [`pkg/backend.Renders()`](https://github.com/bookingcom/carbonapi/blob/b5dd1535f91fd0e864c5bebf0e652ba0a98c8dfd/pkg/backend/rpc.go#L44). It calls a bunch of `pkb/backend/Backend.Render()` in a goroutine and then waits in a loop for each of their result/errors. From what I can tell everything that happens inside `Backend.Render()` is governed by context cancellation except for a very minor edge case. 

How does the delay of cancellation manifest? Is there an easy way to provoke the issue?

gunnihinn:
> How does the delay of cancellation manifest? Is there an easy way to
> provoke the issue?

I'm not sure what you mean by the delay of cancellation. If you're
wondering how `backend.Renders` deals with context cancellation, you can
setup `backend/mock` backends that cancel the context they're given.

If you're asking more about what the `Render` method of `backend/net`
does when a context gets cancelled, you should also be able to setup an
ad-hoc server that stalls until a context is cancelled, either with
`net/http/httptest` or just `net/http`.

I wrote this back in the day, so if it's awful you can curse me. :)

grzkv:
@jwkohnen We definitely know that context cancellations fail from time-to-time. We can clearly see the cases when the requests do not return after the request timeout is exceeded (which causes a context cancel). The exact diagnosis of this is what this issue is about.

Right now I do not know how to reproduce this. We can try to look together and figure it out.

grzkv:
This should be resolved by the processing redesign.

PR #193, opened 2019-09-09
WIP: Have immediate context cancellations in all handlers #149

jwkohnen:


pkg/backend/net/net.go
```diff
@@ -237,7 +237,10 @@ func (b Backend) do(ctx context.Context, trace types.Trace, req *http.Request) (
 
 	go func() {
 		resp, err := b.client.Do(req)
-		ch <- requestRes{resp: resp, err: err}
+		select {
+		case ch <- requestRes{resp: resp, err: err}:
```
gunnihinn: Does this select then change the behavior of the one below in a predictable way? IIRC the runtime doesn't make guarantees about which channel it reads from if more than one are ready (double-IIRC the limiter code assumes that).

In any case, if the ctx cancels then the original request has gotten/is getting/will get an error response, so it matters less exactly what we do to the response here than that we clean up any goroutines and resources we've taken out.

jwkohnen: But there is a race: Given that ctx is canceled, both the reader of the `requestRes` from `ch` below and the `b.client.Do` stop doing what they are doing. That is: the reader below may return upon reading on `ctx.Done()` and will not try to read the `requestRes`. At the same time `b.client.Do` terminates with a `context.Canceled`, tries to send that into `ch` which has no reader anymore and block forever. Thence there is a goroutine leak. 

A possible regression might be that we lose an error that appears right before a timeout by coming late to the race. That is an unlikely coincidence and the lost error may not be worth reporting then?

I have the feeling we're saying the same thing are we?

grzkv:
#149

grzkv:
Could we keep the work on go modules separate, please? It creates very many changes, so context cancel issue becomes hard to follow.

jwkohnen:
I've create a separate, clean PR in favour of this one: #199.

Issue #189, opened 2019-09-09
Move from gometalinter to golandci-lint


jwkohnen:
Gometalinter has been deprecated since 2019-04-07, the community consensus is to use golangci-lint now.

jwkohnen:
fun fact: I wrote this is issue while waiting for gometalinter to terminate. I am 2 minutes in...

grzkv:
Would you like to pick this one up?

jwkohnen:
Sure.

PR #194, opened 2019-09-09
lint: use golangci-lint, initial project config

jwkohnen:
I tried to create a meaningful project local config in .golangci.yml, but of
course this is a matter of taste and debate.  Most of the settings come from
the default golangci-lint config file.

As of now golangci-lint finds quite a bunch of issues.  IMHO they're all worth
investigating.  I have explicitly exluded "composite literal uses unkeyed
fields" in test files as this style is used widely in the table driven tests.

The invocation via make shows only issues that are introduced since the
merge-base of this commit on the master branch in order to make the
introduction painless in CI pipelines etc.

/fixes: #189

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,206 @@
+run:
+  deadline: 1m
+  tests: true
+  build-tags: [ cairo ]
+  skip-dirs: []
+  skip-files:
+    - ".*\\.pb\\.go$"
+  # activate after switching to go modules (#131, #190)
+  # modules-download-mode: vendor
+
+# output configuration options
+output:
+  # colored-line-number|line-number|json|tab|checkstyle|code-climate, default is "colored-line-number"
+  format: colored-line-number
+
+  # print lines of code with issue, default is true
+  print-issued-lines: true
+
+  # print linter name in the end of issue text, default is true
+  print-linter-name: true
+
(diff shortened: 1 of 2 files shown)


grzkv:
I am not sure we can include this in Travis CI now since linting will always fail. Successful Travis run is currently required for PR merge.

jwkohnen:
Yeah, I figured after seeing it fail, because the binary is not available. We could integrate it into GitHub via https://golangci.com/ which may be favourable anyway. I could move the `--new-from-rev` into the `.golangci.yml` and create a special Makefile target which "reports all the gore".

grzkv:
Let's just remove it from the Travis for now and make an issue to do that in the future. How about that?

grzkv:
> a special Makefile target which "reports all the gore"
This makes sense.

ksurent:
golangci-lint vendors a fairly old version of staticcheck (from Jan 2019). I think we should disable the vendored version and run the latest standalone. If we do that, we should also revisit other linters enabled in golangci-lint because they may overlap.

jwkohnen:
squashed. There are now two make targets: lint and lint-all. The first uses the current merge-base as baseline, the latter shows everything. Travis does not use golangci-lint, yet. I'll create follow-up issues for Travis and for using a newer staticcheck.

PR #195, opened 2019-09-09
Reanimate expression parser fuzzer

ksurent:
This PR re–introduces go-fuzz integration and also fixes a bunch of crashers found by it.


grzkv:
I am not sure *ErrUnexpectedCharacter* is the error type we need.

PR #198, opened 2019-09-10
Some unit tests

kostty:


grzkv:
Looks good! Would be even better with [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests). You could take this into account for the future contributions.

PR #199, opened 2019-09-10
backend/net: terminate goroutine on canceled outbound request

jwkohnen:
The parent goroutine reads on the request channel, but stops doing so in a
select case on ctx.Done().  The goroutine that does the outbound request may
want to send the context.Canceled error into that abandoned channel and block
forever.  This change releases the goroutine doing the outbound request on
context cancellation.


jwkohnen:
While searching for a way to debug ctx cancellation I saw some test code and quickly made a proof:

```
jk@artip:~/legacygopath/carbonapi/src/github.com/bookingcom/carbonapi$ git diff
diff --git a/pkg/backend/net/net.go b/pkg/backend/net/net.go
index 10aecc2..0a6f22a 100644
--- a/pkg/backend/net/net.go
+++ b/pkg/backend/net/net.go
@@ -11,6 +11,7 @@ import (
        "net/url"
        "strconv"
        "strings"
+       "sync"
        "time"

        "github.com/bookingcom/carbonapi/pkg/types"
@@ -235,7 +236,11 @@ func (b Backend) do(ctx context.Context, trace types.Trace, req *http.Request) (
        ch := make(chan requestRes)
        t0 := time.Now()

+       var wg sync.WaitGroup
+       wg.Add(1)
+       defer wg.Wait()
        go func() {
+               defer wg.Done()
                resp, err := b.client.Do(req)
                ch <- requestRes{resp: resp, err: err}
        }()
```

... with this hack the do() method blocks in `TestCallTimeout` some of the time. Felt like between 25% and 50%, depending on luck.

The code with the select and the wait group runs okay with `go test --count=10000 --run TestCallTimeout`.

avereha:
This change is fixing the leaking of routines.
Looking at the enclosing function I wonder if we need to make the http request in a separate routine in the first place. If we don't make new routine, we don't have what to leak :)
Another issue with that do() is that it seems to receive two contexts, one as parameter and one inside the req.

grzkv:
Closing as an alternative to #213 that is preferred.

Issue #171, opened 2019-06-17
Add strict from and until times verification in api


grzkv:
Currently, `api` accepts some times that are either too liberal or just incorrect. In this case,  the defaults are applied that either are not obvious or include timeframe that starts "at the beginning of time" (for Unix).

E.g.
```
http://localhost:8081/render?target=a.a.a&from=+5m (future)
http://localhost:8081/render?target=a.a.a&until=+5m (from implied to come since forever and into the future)
```
There are more similar cases.
 
This is closely related to #160

grzkv:
@KozzyKoder You may be interested in this one.

azhiltsov:
Also would be nice to prevent swap of from and until
from go-carbon logs:
Invalid time interval: from time '1562536741' is after until time '1561932061'
--

bom-d-van:
the issue is resolved by https://github.com/bookingcom/carbonapi/pull/200 and https://github.com/bookingcom/carbonapi/pull/341 now.

PR #200, opened 2019-09-10
Added more strict validation on from/until parameters in api

KozzyKoder:
Hello, this PR relates to this issue:
https://github.com/bookingcom/carbonapi/issues/171

Idea is to find to extend test coverage and find permissive validation rules and restrict them by returning an error instead of default value.


app/carbonapi/http_handlers.go
```diff
@@ -173,7 +173,7 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request) {
 		logAsError = true
 		return
 	}
-	if form.from32 == form.until32 {
+	if form.from32 >= form.until32 {
```
azhiltsov: Error message in the next line need to be adjusted as well in order to represent this code change

azhiltsov: For simplicity you can use "Invalid or empty time range"  but if this message is a part of HTTP error response than it make sense to be more verbose. 
Excessive verbosity in the logs isn't needed

KozzyKoder: Added more precise error text there

KozzyKoder:
@azhiltsov also some things might be good to think about:

1) checked how wrong intervals processed at graphite python. Seems it accepts now only from=until, but also until < from. In latter case it looks like it swaps from <-> until. Not intuitive behaviour, is it fine that we restrict this behaviour?

2) added more restrictions on formats itself:

a) "hh:mm 19940812", "12:30:00 19940812", "12:mm 19940812" - similar inputs with incorrect time now returns an error instead 00:00 in current implementation

b) "-something" - completely incorrect relative times also returns error

c)"17:04 19940812 1001" - more components than required now also results in error

d)"12:30 08-15-06" - date with wrong delimiter returns error

e)"08/15/06 12:30" - swapped date and time also considered as error

f)"+5m" - relative dates in future not allowed also

--------------
previously cases described above were returning default value(from - 24hours back, until -now)

What do you think about these cases? Shall we relax some of these restrictions?

azhiltsov:
Original goal of this project was to implement the same functionality as graphite-web. 
We were pursuing this goal for a while by even reproducing some behaviors which might be perceived as buggy. 
I do not have strong opinion about behavior we expect from time parsing. The only thing we should cut is ambiguous parameters. Grafana for instance performs conversion internally and uses epoch notation, so using extending parsing is mostly targeting human interaction. Most probably graphite-web is using some of human readable notations, and I would call @deniszh to let us know if we are about to brake it. 
@gksinghjsr and @grzkv what is your take on this?
I would propose to use "keep what make sense"  approach and what does not (but presented in graphite-web) can be opt_in with compatibility config flag.
My vision that carbonapi shouldn't lock-in on whisper as back-end and if so then in some other back-ends queries in the future for instance could make sense.

grzkv:
I don't think we should closely follow python implementation in everything, only in parts where it makes sense. The strong criterion is: Grafana should always work well. As @azhiltsov has described, it uses Unix time so it should be ok.

I think we should drop the behaviour that is confusing or just wrong. Specific examples:

* `until==from` does not make any sense
* `until < from` is the clearly a user input mistake, treating this as an OK request is _very_ misleading
* `+5m` is user input error
* `-something_that_is_not_time` is user error

@KozzyKoder To sum, I agree with the points *a, b, c, e, f*. I think, *d* is debatable, but I rather agree as well.

date/date.go
```diff
@@ -9,7 +9,10 @@ import (
 	"github.com/bookingcom/carbonapi/pkg/parser"
 )
 
-var errBadTime = errors.New("bad time")
+var errBadTime = errors.New("Time has incorrect format")
```
ksurent: By convention, error strings start with a lowercase letter.

grzkv:
Hi all. There are a lot of valuable contributions in this PR. How about we move this completion?

grzkv:
@azhiltsov @avereha @grlvrl @KozzyKoder Let's finalize this. We either move on or close.

Taking into account the age and controversy, I suggest we close this.

PR #202, opened 2019-09-10
Fix linearRegression index out of range panic

paskal:
I found out following:
`linearRegression(test.test)` : returns metrics
`linearRegression(test.test, "some")` : returns `Internal Server Error (500) Details: all targets failed with mixed errrors; merged errs: (expression eval failed: panic duting expr eval: runtime error: index out of range, )`
`linearRegression(test.test, "some", "some")` : returns metrics

This PR fixes the problem.


grzkv:
Good one. Thx.

PR #207, opened 2019-09-10
Fixes #119. Update http handlers for api and zipper to use gorilla/mux

### LINKED ISSUE #119: Improve HTTP muxer
There are several problems with current HTTP router (muxer):

- [x] some requests are served only when ending with a `/`, e.g. in `zipper` */metrics/find* does not work while */metrics/find/* does
- [x] routes have to be duplicated to accommodate for the previous point
- [ ] router and app code could be divided into separate files in `zipper`
- [x] a `302` is returned when incorrect endpoint is hit instead of `404`

We should also consider moving to [gorilla/mux](https://github.com/gorilla/mux) muxer.

- [x] Move to gorilla/mux
- [x] Routing tests also need to be added.

anjali-sharma:
This provides functionality to add implicit redirect for endpoints with
optional forward slash and avoid adding redundant endpoints

Update:
* Add middleware to trim trailing slash from the request path (& unit tests)
* Update app tests to include testing for routes
* Refactor app tests to run subtests in parallel
* Refactor app tests to use `httptest` package for creating test requests


azhiltsov:
Since this is changing the http library we are using for a while I would like to understand a performance implications of this change.

anjali-sharma:
@azhiltsov Any suggestions which tests to perform around this?

anjali-sharma:
Also, these failed build should be fixed once the switch to go modules (https://github.com/bookingcom/carbonapi/pull/190) is merged.

anjali-sharma:
As per this benchmark (https://github.com/julienschmidt/go-http-routing-benchmark), we can try using https://github.com/julienschmidt/httprouter which seems to have the best performance vs memory consumption. Though, it doesn't seem to be maintained that frequently.

grzkv:
@azhiltsov Performance is not a concern here. Here's a [benchmark](https://gist.github.com/pkieltyka/123032f12052520aaccab752bd3e78cc#file-gistfile1-txt-L100) (can't say it's very authoritative though). Even in the very worst case, we are looking at 1ms delay max. My guess is that delay will be in microseconds.

grzkv:
Regarding the muxer to use: I'd go with `gorilla/mux`. It has the most established reputation.

grzkv:
I think, `301`s is exactly what we are trying to fix with this PRs.

grzkv:
I would be nice to have routing tests with this PR.

Civil:
@grzkv main problem with 301s is what jwkohnen mentioned.

Grafana uses POST requests to send it's requests (as this allows arbitrary large requests).

If webserver replies with 301, the behavior of most of the HTTP Clients (including browsers) is to change the request to GET and drop the body[1].

So your request to `/render` with body `target=foo` will suddenly become just a `/render` with no parameters.

The correct behavior would be to return `308` instead[1], but then you might still encounter some HTTP Clients that won't understand 308. That's the reason why gorilla/mux doesn't do it by default[3] and that was original reason for having same handler function defined for both URL with and without `/` in the end.

[1] https://golang.org/pkg/net/http/#Client.Do - paragraph that starts with "If the server replies with a redirect"
[2] https://tools.ietf.org/html/rfc7538
[3] https://github.com/gorilla/mux/issues/285

azhiltsov:
> I think, `301`s is exactly what we are trying to fix with this PRs.

$ curl -v 127.0.0.1:8091/render/?format=json
> GET /render/?format=json HTTP/1.1
< HTTP/1.1 200 OK

$ curl -v 127.0.0.1:8091/render?format=json
> GET /render?format=json HTTP/1.1
< HTTP/1.1 200 OK

grzkv:
Hey all! Let's move this PR to some final conclustion.

grzkv:
@avereha Can you please give some feedback?

avereha:
Graphite-web seem to be doing the same trick with "/?", for example:

https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/urls.py

I would look there for other examples/URLs and make sure we don't diverge too much.

Another approach that I noticed was to add middleware that removes tailing shashes. Not sure if this will add too many cases that would be different from graphite-web. Or we could make small muxer that wraps original muxer and supports registering two paths.
I think all approaches would work.

No matter what way we choose, I would still add tests for the next time we want to replace the muxer.

app/carbonapi/app_test.go
```diff
@@ -322,3 +328,14 @@ func TestInfoHandler(t *testing.T) {
...
+func setupTestServer(testHandler http.Handler) *httptest.Server {
+	testListener, err := net.Listen("tcp", "127.0.0.1:8080")
+	if err != nil {
+		log.Fatal(err)
+	}
+	ts := httptest.NewUnstartedServer(testHandler)
+	ts.Listener.Close()
+	ts.Listener = testListener
+	return ts
+}
```
grzkv: End file with newline.
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

app/carbonapi/routes.go
```diff
@@ -4,57 +4,64 @@ import (
...
+	return routeMiddleware(r)
 }
+
+// routeHelper formats the route using regex to accept optional trailing slash
+func routeMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		r.URL.Path = strings.TrimSuffix(r.URL.Path, "/")
+		next.ServeHTTP(w, r)
+	})
+}
```
grzkv: End file with newline.
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

app/carbonapi/routes_test.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func testRoutingForPath(t *testing.T, path string, router http.Handler) {
+	req := httptest.NewRequest("GET", path, nil)
+	rr := httptest.NewRecorder()
+	router.ServeHTTP(rr, req)
+	if rr.Code == http.StatusNotFound {
+		t.Errorf("Failed to route path: %s", path)
+	}
+}
```
grzkv: End file with newline.
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

app/carbonapi/app_test.go
```diff
@@ -171,17 +182,9 @@ func setUpTestConfig() *App {
...
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	return req
-}
-
-func TestRenderHandler(t *testing.T) {
-	req := setUpRequest(t, "/render/?target=fallbackSeries(foo.bar,foo.baz)&from=-10minutes&format=json&noCache=1")
+func testRenderHandler(t *testing.T) {
```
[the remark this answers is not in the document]
anjali-sharma: These are local methods (not to be executed using `go test` directly). Will rename them to avoid confusing them with unit tests.

grzkv:
@azhiltsov There should be not `301` redirect responses anymore. Could you please re-evaluate your review?

avereha:
The integration test is failing because go-carbon got moved to go-graphite. Will try to fix this is a separate PR because it is not related to this change.

Tested the latest version with production traffic and found there was no 3xx answer, which means we are handling tailing / OK.
Issues that I found:
 - curl -L -v localhost:8091/ goes in infinite redirect loop: curl: (47) Maximum (50) redirects followed 
 - curl -L -v localhost:8091/qwfqewf answers with 404(better than before where we answered with 200). But the usage body is missing.

grzkv:
@avereha @anjali-sharma Is this a WIP or ready for review?

anjali-sharma:
@grzkv It's a WIP. I'm looking into the issues reported by @avereha.

anjali-sharma:
@avereha Added fix for the edge case (thanks for reporting!) and test for it. Also refactored the tests a bit to handle test App instance setup for the package.

avereha:
Tested again and everything works as expected.
Again, there was no 3xx answer.

avereha:
Thank you @anjali-sharma !

PR #208, opened 2019-09-11
Add integration tests to the CI config

kostty:


tests/system_test.sh
```diff
@@ -6,25 +6,48 @@
...
+
+echo "travis_fold:start:docker-up"
+echo -e "\e[33;1mStarting containers\e[0m"
 docker-compose up -d || { printf ">>> ERROR: Launching containers failed"; exit 2; }
+echo # `echo` to work around formatting issues after `docker-compose`
+echo "travis_fold:end:docker-up"
 
 # This is guestimate for the time when the system goes up and operational.
 # Doing this other ways is much more complex. May need to be increased on slow boxes.
 sleep 5
```
paskal: Is it possible to get rid of that sleep?

kostty: Indeed, I couldn't do it quickly before. Will look into it today

grzkv: I could appreciate if could throw it out. Probably, you could send a simple request repeatedly to eval if the system is up, but this adds a lot of complexity. Other ideas?

tests/system_test.sh
```diff
@@ -6,24 +6,48 @@
...
-echo "test.test.test 123456 $(date +%s)" | nc -q 1 localhost 2003 || { printf ">>> ERROR: Putting data in failed"; exit 2; }
+METRIC_NAME="some.test.metric"
+TEST_VALUE=123456
+
+echo "travis_fold:start:testing"
+echo -e "\e[33;1mStoring and retrieving datapoints\e[0m"
+echo "$METRIC_NAME $TEST_VALUE $(date +%s)" | nc -q 1 localhost 2003 || { printf ">>> ERROR: Putting data in failed"; exit 2; }
 
 # wait to make sure data got there
 sleep 2
```
grzkv: This `sleep` problem is similar. Since this is TCP, there is no way to know that the record got there.

One alternative is to repeat the check for e.g. 2 seconds until it succeeds and remove this sleep. This would also introduce a delay but will increase the complexity as well.

tests/system_test.sh
```diff
@@ -6,24 +6,48 @@
...
+echo "travis_fold:start:docker-build"
+echo -e "\e[33;1mBuild docker images\e[0m"
 docker-compose build || { printf ">>> ERROR: Docker build failed"; exit 2; }
+echo "travis_fold:end:docker-build"
+
+echo "travis_fold:start:docker-up"
+echo -e "\e[33;1mStarting containers\e[0m"
 docker-compose up -d || { printf ">>> ERROR: Launching containers failed"; exit 2; }
+echo # `echo` to work around formatting issues after `docker-compose`
+echo "travis_fold:end:docker-up"
```
grzkv: If one runs this script separately from Travis CI, this will look weird.

grzkv:
@kostty AFAIK this will run all the time. If this will slow down the pipeline significantly, we will have to make it run less frequently.

kostty:
> @kostty AFAIK this will run all the time. If this will slow down the pipeline significantly, we will have to make it run less frequently.

Yep. So far the impact is pretty insignificant, the [run times](https://travis-ci.com/bookingcom/carbonapi/builds/127015693) are 2m17s for unit tests and 3m10s for the system test. Since the jobs run in parallel, the overall time only increased by ~1m.

grzkv:
@kostty Thanks for the details! We will merge this shortly.

kostty:
The change suggestions from this PR have been moved to a [new issue](https://github.com/bookingcom/carbonapi/issues/210).

PR #211, opened 2019-09-18
Gysingh/render sli prom

gksinghjsr:
Adding prometheus metric with linear buckets tracking 'simple' requests. Needed for more granular and uniform buckets, in comparison to exponential buckets.


grzkv:
LGTM

PR #213, opened 2019-09-24
backend/net: terminate goroutine on canceled outbound request

jwkohnen:
The parent goroutine reads on the request channel, but stops doing so in
a select case on ctx.Done().  The goroutine that does the outbound
request may want to send the context.Canceled error into that abandoned
channel and block forever.

This change releases the goroutine doing the outbound request on context
cancellation by ensuring a successful send into a buffered channel.  It
is an alternate solution to #199.


paskal:
I'd prefer #199 rather than this one.

grzkv:
@avereha Can you please share the feedback on this one? Especially as compared to #199

avereha:
I posted another comment on #199.
Comparing the fixes, I think any would work and don't have any strong preference for one vs. another.

jwkohnen:
I consider this the "canonical solution" to this sort of problem and prefer this over #199.

grzkv:
Let's merge this one.

PR #216, opened 2019-10-18
metrics: adding linear bucket histogram for find

azhiltsov:


cfg/common_test.go
```diff
@@ -53,7 +53,10 @@ monitoring:
         start: 0.06
         bucketsNum: 60
         bucketSize: 3
-
+	findDurationLinHistogram:
+        start: 0.5
+        bucketsNum: 20
+        bucketSize: 0.5
```
grzkv: It would be good if the numbers in the test were different from defaults. Then we can actually test the parsing.

config/carbonapi.yaml
```diff
@@ -136,3 +136,11 @@ monitoring:
         start: 0.0
         bucketsNum: 40
         bucketSize: 0.05
+    findDurationExpHistogram:
+        start: 0.06
+        bucketsNum: 30
+        bucketSize: 3
```
grzkv: Multiples of 3 are usually not very comfortable. Are there any particular reasons to use multiples of 3?

PR #220, opened 2019-11-26
Add timeLag function for computing a time lag

justpresident:
Given two monotonically non-decreasing time series (series1,series2), for every datapoint of series1 it computes the time difference required for series1 metric to achieve value of series2.

The easiest way to vizualize its correct work is to produce some ramdom
monotonnically increasing metric. Let's call it 'local.random.diceroll'
and then render the following url: render?target=timeLag(timeStack(local.random.diceroll,'1min',0, 7), local.random.diceroll)&from=-10m
Even though metric values are random, the resulting graph should look like horizontal lines precisely matching time offsets passed to timeStack function

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -167,6 +167,7 @@ _When `format=png`_ (default if not specified)
 - powSeriesLists
 - removeZeroSeries
 - stdev
+- timeLag
 - tukeyAbove
 - tukeyBelow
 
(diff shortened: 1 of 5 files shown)


azhiltsov:
Please update this https://github.com/bookingcom/carbonapi/blob/master/COMPATIBILITY.md

azhiltsov:
I am fine with the name, but for sake of potential future back porting to the python graphite, @deniszh do you have an opinion on the name for this function?

gunnihinn:
What is `timeLag(s1, s2)` where
```
s1(t) = 0
s2(t) = 1
```
for all `t`?

justpresident:
> What is `timeLag(s1, s2)` where
> 
> ```
> s1(t) = 0
> s2(t) = 1
> ```
> 
> for all `t`?
Good question. it is undef for all datapoints

justpresident:
> Please update this https://github.com/bookingcom/carbonapi/blob/master/COMPATIBILITY.md

done

COMPATIBILITY.md
```diff
@@ -288,8 +290,10 @@ _When `format=png`_ (default if not specified)
 | summarize(seriesList, intervalString, func='sum', alignToFrom=False)      |
 | threshold(value, label=None, color=None)                                  |
 | timeFunction(name, step=60), Short Alias: time()                          |
+| timeLagSeries(consumeMaxOffsetSeries, produceMaxOffsetSeries)             |
+| timeLagSeriesLists(consumeMaxOffsetSeriesLists, produceMaxOffsetSeriesLists) |
```
grzkv: I don't see the implementation of this function.

justpresident: both functions defined in expr/functions/timeLag. I could've done similar to the way holtwinters logic is shared between two functions, but since they have absolutely the same logic I did it the same way as expr/functions/seriesList

PR #227, opened 2020-01-06
Instrumenting zipper with per cluster/host performance metrics. 

grlvrl:
Instrumenting zipper with per cluster/host performance metrics. 
We first changed configuration to include cluster info, then decorated zipper prometheus metrics with cluster information.

### DIFF
--- a/app/carbonzipper/app.go
+++ b/app/carbonzipper/app.go
@@ -199,8 +199,9 @@ func initBackends(config cfg.Zipper, logger *zap.Logger) ([]backend.Backend, err
 		}).DialContext,
 	}
 
-	backends := make([]backend.Backend, 0, len(config.GetBackends()))
-	for _, host := range config.GetBackends() {
+	configBackendList := config.GetBackends()
+	backends := make([]backend.Backend, 0, len(configBackendList))
+	for _, host := range configBackendList {
 		cluster, _ := config.ClusterOfBackend(host)
 		b, err := bnet.New(bnet.Config{
 			Address:            host,
--- a/app/carbonzipper/app.go
+++ b/app/carbonzipper/app.go
@@ -199,10 +199,12 @@ func initBackends(config cfg.Zipper, logger *zap.Logger) ([]backend.Backend, err
 		}).DialContext,
 	}
 
-	backends := make([]backend.Backend, 0, len(config.Backends))
-	for _, host := range config.Backends {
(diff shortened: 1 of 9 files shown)


app/carbonzipper/app.go
```diff
@@ -199,10 +199,12 @@ func initBackends(config cfg.Zipper, logger *zap.Logger) ([]backend.Backend, err
 		}).DialContext,
 	}
 
-	backends := make([]backend.Backend, 0, len(config.Backends))
-	for _, host := range config.Backends {
+	backends := make([]backend.Backend, 0, len(config.GetBackends()))
```
gksinghjsr: The result of this can be cached to a variable, instead of being inline, and used in the next statement.

pkg/backend/net/net.go
```diff
@@ -270,7 +274,7 @@ func (b Backend) do(ctx context.Context, trace types.Trace, req *http.Request) (
 		return res.resp.Header.Get("Content-Type"), body, nil
 
 	case <-ctx.Done():
-		trace.ObserveOutDuration(time.Now().Sub(t0))
+		trace.ObserveOutDuration(time.Now().Sub(t0), b.cluster)
```
grzkv: I am not sure we should count cancelled and timed-out requests. (And these are the ones that fall in here)

cfg/common.go
```diff
@@ -152,6 +155,43 @@ type Common struct {
...
+		}
+	}
+
+	for _, backend := range common.Backends {
+		if backend == address {
+			return "", nil
+		}
+	}
+
+	return "", errors.Errorf("Couldn't find cluster for '%s'", address)
```
grzkv: It is better not to have capitals in errors. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

app/carbonzipper/app.go
grzkv: +1 AFAIU we are talking about `GetBackends()`.

grlvrl: I have updated and committed the change.

pkg/backend/net/net.go
grlvrl: @gksinghjsr , @azhiltsov , @avereha - what do you think on this comment?

cfg/common.go
grlvrl: I agree with you, but all the instances of "errors.Errorf" in the project start with capitals.

PR #231, opened 2020-01-24
add missing config file for zipper

paskal:
$subj


grzkv:
This is designed to work with docker-compose and not in isolation. The config file is in place https://github.com/bookingcom/carbonapi/blob/master/docker-compose.yaml#L26

Issue #232, opened 2020-01-27
Fix auto-complete suggestions for groupByNode


grzkv:
`groupByNode(node, func)` works for func = *sum* and func = *max*, but does not work for average if one uses the autocomplete suggestion. This is because we expect *avg* but suggest *average*. 

There are two ways to resolve this:
1. Fix suggestions only by changing it from *average* to *avg*.
2. Fix suggestions and allow for both *avg* and *average* to be used. This will provide full backward-compatibility with graphite-web (see https://github.com/bookingcom/carbonapi/blob/master/expr/functions/groupByNode/function.go#L141)

PR #234, opened 2020-01-27
auto complete group by node

grlvrl:
Fix auto-complete suggestions for groupByNode #232

https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.averageSeries states that averageSeries is an alias for aggregate with aggregation average. Therefore "average" method is also added to method list of github.com/bookingcom/carbonapi/expr/averageSeries

Same applied for "diff" and "stddev", which are also alias for "diffSeries" and "stddevSeries" respectively according to graphite documentation.


grzkv:
It looks good, but I see too many commits in this PR. Probably forked from older place in `master`. Makes sense to rebase and force-push maybe.

grlvrl:
I think I fixed the issue of too many commits.

Issue #101, opened 2019-03-19
moving* functions shifting results to the left query boundary


azhiltsov:
When query executed on a whisper archive crossing it left boundary the results of movingMedian movingSum movingMin movingMax movingMax (might be more)
are shifted to the left query boundary:
Lets say you have a whisper archive with a single archive storing 2 days of secondly data:
```json
{
"10.10.10.10:8080": {
"name": "secondly.views_sum",
"aggregationMethod": "Sum",
"maxRetention": 172800,
"retentions": [
{
"secondsPerPoint": 1,
"numberOfPoints": 172800
}
]
}
```
you querying it on the range from=now-3d until=now
you getting 2 days of data, padded with nulls from the LEFT
applying moving* function on top of it will shift result to the left, padding it with nulls from the RIGHT
![image](https://user-images.githubusercontent.com/8709761/54609536-7f4cee80-4a53-11e9-87b4-32c499dffd85.png)

This issue does not exist in ewma function for instance

PR #235, opened 2020-01-27
#101 Fix the issue with moving* and non-available data points on the left + fix NaN treated like 0

nvbn:
## Shift problem

The main difference with `ewma` was that `moving*` was setting `startTime` and `endTime` form the "query" to the result.

image

After the changes with two entries at 1580225574 and 1580225575 with `from=1580225568` the result isn't shifted to left:
```json
[
  {
    "target":"movingSum(new.file,2)",
    "datapoints":[
      [
        null,
        1580225574
      ],
      [
        30,
        1580225575
      ],
      [
        null,
        1580225576
      ],
      [
        null,
        1580225577
      ],
      [
        null,
        1580225578
      ],
      [
        null,
        1580225579
      ]
    ]
  }
]
```

## NaNs problem
With the data available fro 09:00 till 10:00 `movingSum` before the changes:
image

And after the changes (no zeros):
image

The data set - https://gist.github.com/nvbn/037c1b4bf502955c072280761fc74c1d


grzkv:
@nvbn Can you please share a bit of background and explanation about the root cause of the bug? Thanks!

grzkv:
@nvbn Also, can you please share the results of your tests before and after the bug fix?

nvbn:
@grzkv  the root cause is the `StartTime` and `StopTime` from a metric [was overwritten by `from` and `until` from the query](https://github.com/bookingcom/carbonapi/blob/2feef0c7d3d9fed20da56e43d63933f96c9ea265/expr/functions/moving/function.go#L90).

Steps to reproduce:
* `docker-compose up`
* `docker-compose exec go-carbon mkdir -p /data/graphite/whisper/new`
* download [whisper file limited to 10 entries](https://github.com/bookingcom/carbonapi/files/4127641/file.txt) and rename it to `file.wsp
* `docker cp file.wsp carbonapi_go-carbon_1:/data/graphite/whisper/new`
* write some data:
```bash
export TS=$(date +%s)
echo "new.file 15 $TS"  |  nc -c localhost 2003
echo "first entry $TS"
sleep 1
export TS=$(date +%s)
echo "new.file 15 $TS" | nc -c localhost 2003
echo "second entry $TS"
curl http://localhost:8081/render\?target\=movingSum\(new.file,%201\)\&format\=json\&from\=$(python -c "print($TS-15)")
```
* look at the output:
```
first entry 1580297543
second entry 1580297544
{"target":"movingSum(new.file,1)","datapoints":[[null,1580297529],[0,1580297530],[0,1580297531],[0,1580297532],[0,1580297533],[0,1580297534],[0,1580297535],[0,1580297536],[15,1580297537],[15,1580297538]]}]
```
* see that the datapoints are shifted to the left:

entry | datapoint
--- | ---
1580297543 | 1580297537
1580297544 | 1580297538

**After the changes** output is:
```
first entry 1580298244
second entry 1580298245
[{"target":"movingSum(new.file,1)","datapoints":[[null,1580298236],[null,1580298237],[null,1580298238],[null,1580298239],[null,1580298240],[null,1580298241],[null,1580298242],[null,1580298243],[15,1580298244],[15,1580298245]]}]
```
The values are the same (and also no zeros):

entry | datapoint
--- | ---
1580298244 | 1580298244
1580298245 | 1580298245

grzkv:
I am not quite able to replicate the test in docker-compose following the steps described. I get a different points alignment.

Also, in your case you still get zeros when window size is 1. I guess, in that case zeros must in fact be nulls. Am I getting it right?

nvbn:
@grzkv did you use the same whisper file?

> Also, in your case you still get zeros when window size is 1. I guess, in that case zeros must in fact be nulls. Am I getting it right?

There are no zeros after the change, only nulls.

grzkv:
I cannot fully repeat the result, but this is not strong evidence that the solution does not work. It can be due to how `go-carbon` treats Whisper files with low data-points as time passes. Approved, will double-check during rollout.

Issue #233, opened 2020-01-27
Replace carbon-c-relay with nanotube in docker-compose and related tests


grzkv:
The `carbon-c-relay` needs to be updated with `nanotube` in docker-compose and tests on top. The `carbon-c-relay` config needs to be replicated and cleaned-up.

PR #236, opened 2020-01-28
Add nanotube as an option on docker-compose

laq:
Include nanotube as a relay option. 
crelay is still available as a commented out option.

See https://github.com/bookingcom/carbonapi/issues/233

For this to work properly the nanotube import paths have to be fixed. https://github.com/bookingcom/nanotube/pull/9


grzkv:
I think it makes sense to clean up `crelay`. Also, did you run some tests to verify if everything works?

docker/nanotube/nanotube.conf
```diff
@@ -0,0 +1,9 @@
+cluster a34000
```
grzkv: I am not sure I understand the purpose of this file. It's a crelay config but looks like nanotube one. Could you please explain?

docker/nanotube/config/config.toml
```diff
@@ -0,0 +1,53 @@
+# The settings are listed in this file are with their default values.
```
grzkv: I think these comments are applicable mostly within the `nanotube` repo.

docker/nanotube/nanotube.conf
laq: It was a left over from coping crelay config. Removed.

docker/nanotube/config/config.toml
laq: Just removed them

PR #240, opened 2020-02-17
[WIP] Added new mode to use memcached. 

grzkv:
Now it is possible to use a set of Memcached replicas, not only a ring.

I have added a switch to the config that allows considering all Memcached instances supplied in config as identical replicas. The workings are as follows:

### Write

Write is sent to all instances equally. The result of writing is ignored.

### Read

Read request is sent to all instances. The first successful return is considered the result. This includes not-founds. If all instances return an error, it's an error. There is also a newly introduced timeout parameter which represents the maximum amount of time to wait until responses come in. It is not used in both variations of Memcached clusters.

### DIFF
--- a/app/carbonapi/app.go
+++ b/app/carbonapi/app.go
@@ -207,15 +207,23 @@ func setUpConfig(app *App, logger *zap.Logger) {
 		logger.Info("memcached configured",
 			zap.Strings("servers", app.config.Cache.MemcachedServers),
 		)
-		app.queryCache = cache.NewMemcached("capi", app.config.Cache.MemcachedServers...)
-		app.findCache = cache.NewMemcached("capi", app.config.Cache.MemcachedServers...)
 
-		mcache := app.queryCache.(*cache.MemcachedCache)
+		if app.config.Cache.Replicated {
+			app.queryCache = cache.NewReplicatedMemcached("capi", app.config.Cache.QueryTimeoutMs, app.config.Cache.MemcachedServers...)
+			app.findCache = cache.NewReplicatedMemcached("capi", app.config.Cache.QueryTimeoutMs, app.config.Cache.MemcachedServers...)
 
-		apiMetrics.MemcacheTimeouts = expvar.Func(func() interface{} {
-			return mcache.Timeouts()
-		})
-		expvar.Publish("memcache_timeouts", apiMetrics.MemcacheTimeouts)
+		} else {
+			app.queryCache = cache.NewMemcached("capi", app.config.Cache.QueryTimeoutMs, app.config.Cache.MemcachedServers...)
+			app.findCache = cache.NewMemcached("capi", app.config.Cache.QueryTimeoutMs, app.config.Cache.MemcachedServers...)
+
+			mcache := app.queryCache.(*cache.MemcachedCache)
+
+			// TODO (grzkv) Move to conventional Prom metrics.
+			apiMetrics.MemcacheTimeouts = expvar.Func(func() interface{} {
+				return mcache.Timeouts()
+			})
+			expvar.Publish("memcache_timeouts", apiMetrics.MemcacheTimeouts)
+		}
 
 	case "mem":
 		app.queryCache = cache.NewExpireCache(uint64(app.config.Cache.Size * 1024 * 1024))
--- a/app/carbonapi/app.go
+++ b/app/carbonapi/app.go
@@ -220,7 +220,7 @@ func setUpConfig(app *App, logger *zap.Logger) {
 		expvar.Publish("memcache_timeouts", apiMetrics.MemcacheTimeouts)
 	case "memcacheReplicated":
 		if len(app.config.Cache.MemcachedServers) == 0 {
-			logger.Fatal("replicatred memcache cache requested but no memcache servers provided")
+			logger.Fatal("replicated memcache cache requested but no memcache servers provided")
 		}
 		logger.Info("replicated memcached configured",
 			zap.Strings("servers", app.config.Cache.MemcachedServers))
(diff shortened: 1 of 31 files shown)


cache/cache.go
```diff
@@ -107,3 +117,95 @@ func (m *MemcachedCache) Set(k string, v []byte, expire int32) {
...
+	for _, s := range servers {
+		m.instances = append(m.instances, memcache.New(s))
+	}
+
+	return &m
+}
+
+// Get gets value for the key from the replicated memcached.
+// It sends the request to all replicas and picks the first valid answer
+// (event if it's a not-found) or times out.
```
azhiltsov: would it be hard to wait for first 'HIT' instead of any valid reply?
A do understand that this increasing a waiting time, but we could compensate this by stricter timeouts for cache queries.

grzkv:
No, it’s not hard. But do we actually want it?
Need more data.

cache/cache.go
grzkv: Also, this use case means that we make use of inconsistent data on purpose. I would rather ignore that kind of data if possible.

grzkv: Otherwise, it would be nice to have more input. @avereha @gksinghjsr @grlvrl What do you think?

cache/cache.go
```diff
@@ -107,3 +117,95 @@ func (m *MemcachedCache) Set(k string, v []byte, expire int32) {
...
+func (m *ReplicatedMemcached) Get(k string) ([]byte, error) {
+	resCh := make(chan cacheResponse)
+
```
avereha: Where do we send the requests?

cache/cache.go
```diff
@@ -107,3 +117,95 @@ func (m *MemcachedCache) Set(k string, v []byte, expire int32) {
...
+	return nil, errors.New("all caches failed with errors: " + cacheErrs)
+}
+
+// Set sets the key-value pair for all cache instances.
+func (rm *ReplicatedMemcached) Set(k string, val []byte, expire int32) {
+	key := sha1.Sum([]byte(k))
+	hk := hex.EncodeToString(key[:])
+	for _, m := range rm.instances {
+		go func(k_ string, val_ []byte, expire_ int32, m_ *memcache.Client) {
+			m_.Set(&memcache.Item{
```
avereha: We could add counter here for number of Set errors.

cache/cache.go
```diff
@@ -107,3 +117,95 @@ func (m *MemcachedCache) Set(k string, v []byte, expire int32) {
 func (m *MemcachedCache) Timeouts() uint64 {
 	return atomic.LoadUint64(&m.timeouts)
 }
+
+type ReplicatedMemcached struct {
+	prefix    string
```
avereha: What do we use prefix for?

cfg/api.go
```diff
@@ -105,10 +106,13 @@ type API struct {
...
-	Type              string   `yaml:"type"`
-	Size              int      `yaml:"size_mb"`
-	MemcachedServers  []string `yaml:"memcachedServers"`
-	DefaultTimeoutSec int32    `yaml:"defaultTimeoutSec"`
+	Type             string   `yaml:"type"`
+	Size             int      `yaml:"size_mb"`
+	MemcachedServers []string `yaml:"memcachedServers"`
+	// TODO (grzkv): This looks to be used as expiration time for cache
+	DefaultTimeoutSec int32  `yaml:"defaultTimeoutSec"`
+	Replicated        bool   `yaml:"replicated"`
```
avereha: Would it make sense to define this as a different type of cache("replicated")?
That way, we don't have to think about meaning of other combinations(for example: Type=mem, replicated=true)?

cache/cache.go
avereha: I would keep as simple as possible and test.
I guess if "not found" answers are always faster, then we should see this when we restart random memcache instances.
Worst case I can imagine would be restarting memcache in a loop.

grzkv: @avereha would you jeep the current logic?

cache/cache.go
grzkv: This is a broader issue. We need to add a while set of metrics. In addition to that we also need to change the way the counting is done for old memcached. Currently we use an expvar where Prometheus should’ve been used. I will tackle that in a separate PR.

cache/cache.go
avereha: Yes

grzkv: Let's keep it.

cfg/api.go
grzkv: This is more involved than this. The thing is that the memcache type of cache that already was there has the same problem. You can have type mem + replicas. So, we need to refactor config for it as well. This is not a very small refactoring. I'd do it separately.

grzkv:
@avereha Please have another look.

grzkv:
@azhiltsov added docs

app/carbonapi/app.go
```diff
@@ -207,15 +207,26 @@ func setUpConfig(app *App, logger *zap.Logger) {
...
 		mcache := app.queryCache.(*cache.MemcachedCache)
 
+		// TODO (grzkv) Move to conventional Prom metrics.
 		apiMetrics.MemcacheTimeouts = expvar.Func(func() interface{} {
 			return mcache.Timeouts()
 		})
 		expvar.Publish("memcache_timeouts", apiMetrics.MemcacheTimeouts)
+	case "memcacheReplicated":
+		if len(app.config.Cache.MemcachedServers) == 0 {
+			logger.Fatal("replicatred memcache cache requested but no memcache servers provided")
```
avereha:
```suggestion
			logger.Fatal("replicated memcache cache requested but no memcache servers provided")
```

grzkv:
@azhiltsov Added the docs. Can you please approve. Otherwise, merging is blocked.

PR #244, opened 2020-03-04
Goroutines were blocked if reqeuests were timed out. Fixed.

grzkv:
This fixes a goroutine leak.


avereha:
If this case gets triggered on Timeouts, I think we should set Timeout for memcache library to a lower value than our timeout.
Otherwise, the goroutine will continue running while we stopped reading/caring about the answer.

PR #254, opened 2020-03-31
Add pull request template

avereha:
I'm trying to add a pull request template by following the documentation here:

https://help.github.com/en/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository

That should make it easier for people to follow contribution guidelines.
Let me know what you think about the questions in the template.


grzkv:
@avereha do you want all PRs to follow this?

doc/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+
+## What issue is this change attempting to solve?
+
+## How does this change solve the problem? Why is this the best approach?
+
+## How can we verify this works as expected?
+
+## What feedback would you like, if any?
```
grzkv, line 8: I don't think this is necessary. It's not a must (just as it says).

doc/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+
+## What issue is this change attempting to solve?
+
+## How does this change solve the problem? Why is this the best approach?
+
+## How can we verify this works as expected?
```
grzkv, line 6: I suggest we re-phrase to
> How can we be sure this works as expected?
or 
> How did you check it works as expected?

doc/pull_request_template.md
avereha: Yes, I was thinking it might useful to set expectations.

TBH I didn't use templates before, my guess is that if we set this up when you open a PR you will see this list of items already filled in.
So they are not making anything "required", they are more like a guide. If you want to leave them empty(or remove them) I think it's still possible.

doc/pull_request_template.md
grzkv: @avereha ?

doc/pull_request_template.md
grzkv: then we can remove "if any"
but honestly, I don't think this item is necessary

doc/pull_request_template.md
```diff
@@ -0,0 +1,7 @@
+## What issue is this change attempting to solve?
```
grzkv, line 1: I suggest we also require a link to an issue on GH. This is recommended strategy described in contribution guidelines.

avereha: Do you suggest to change the text?
Or do you know how to update the pull requests template to make some fields "required"?

avereha:
> @avereha do you want all PRs to follow this?

Yes

doc/pull_request_template.md
avereha: Removed!

doc/pull_request_template.md
grzkv: I suggest changing the text. E.g.
> Link to the issue this PR resolves

doc/pull_request_template.md
avereha: Done

doc/pull_request_template.md
avereha: Done

grzkv:
I suggest we move this to the .github directory because it is GitHub-specific.

doc/pull_request_template.md
avereha: Actually, reading https://github.com/bookingcom/carbonapi/blob/master/CONTRIBUTING.md again we have: "Refer to relevant issue if it's there.".
My understanding is that we can have PRs without linked issues. In this case I would use the initial phrasing so we could: 
a) link to issue if it's there
b) explain in the words what is this change about

grzkv: I thought these fields in the PR are optional. So, if we require the issue link, it is by default optional. Isn't it?

avereha: Yes.
But then if there is not issue to link to, we can still have a short explanation, even if it's using words instead of links.

PR #258, opened 2020-05-11
expr: introduce a new func transformNonNull

bom-d-van:
One of the use cause for this func could be used with transformNull to count
changes of metric size:
    sum(transformNull(transformNonNull(ns1.ns2.*.count), 0))

Add a symmetrical new func to transformNull: transformNonNull

There is no alternative function exist in the tool.

There are tests and it's has been tried out on my local env against our production servers.

image

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -170,6 +170,7 @@ _When `format=png`_ (default if not specified)
 - timeLagSeriesLists
 - tukeyAbove
 - tukeyBelow
+- transformNonNull
 
 ## Function short docs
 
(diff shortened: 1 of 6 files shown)


avereha:
Do you need to use this function for something else than counting number of non-null datapoints?
If this is the only use-case, I would use isNonNull: 

https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.isNonNull

bom-d-van:
Yep, for my purpose isNonNull works fine. I think it could be replaced with isNonNull and scale so there seems to be no need to include this function. I will close the MR.
Thanks.

PR #263, opened 2020-07-03
Avereha/tracing

avereha:

Add support for sending traces in Jaeger format from carbonzipper and carbonapi.
Carbonapi is also able to read b3 trace headers that come from grafana.

I'm using this new dependency to send traces:

https://github.com/open-telemetry/opentelemetry-go
and 
https://github.com/open-telemetry/opentelemetry-go-contrib/tree/master/instrumentation/github.com/gorilla/mux
for hooking up to gorilla mux.

Since we don't use gorilla mux yet, information in traces is missing route information, for example:
```
 HTTP POST route not found
```
but this should be fixed after https://github.com/bookingcom/carbonapi/pull/207

Also updated golang version in the travis tests to 1.14 to fix:
```
vendor/go.opentelemetry.io/otel/api/standard/http.go:242:25: undefined: http.StatusEarlyHints
```

I ran multiple tests:
 - tested that we can send traces to opentelemetry-collector and that we are receiving them on the other side.
 - tested on one production server, there was no change in performance:
   - with tracing disabled 
   - with tracing enabled
     - when the collector is working
     - when the collector is not responding/timing out


cmd/carbonapi/main.go
```diff
@@ -64,5 +64,6 @@ func main() {
 	if err != nil {
 		logger.Error("Error initializing app")
 	}
-	app.Start()
+	stop := app.Start()
```
grzkv, line 67: I think we should rename _stop_ -> _flush_

cmd/carbonzipper/main.go
```diff
@@ -78,5 +78,6 @@ func main() {
 	if err != nil {
 		logger.Error("Error initializing app")
 	}
-	app.Start()
+	stop := app.Start()
```
grzkv, line 81: Rename _stop_ -> _flush_

pkg/trace/trace.go
```diff
@@ -0,0 +1,76 @@
...
+	// Grafana propagates traces over b3 headers
+	oldProps := global.Propagators()
+	props := propagation.New(
+		propagation.WithExtractors(propagator),
+		propagation.WithExtractors(oldProps.HTTPExtractors()...),
+		propagation.WithInjectors(oldProps.HTTPInjectors()...),
+	)
+	global.SetPropagators(props)
+
+	return func() {
```
grzkv, line 73: `func() { flush() }` -> `flush`

avereha: Done!

cmd/carbonzipper/main.go
avereha: Done!

cmd/carbonapi/main.go
avereha: Done

pkg/trace/trace.go
avereha: Done

PR #269, opened 2020-07-16
Cleaned dependencies, change build settings.

grzkv:


Makefile
```diff
@@ -34,8 +34,8 @@ nocairo: build
 debug: build
 
 build:
-	$(PKGCONF) $(GO) build $(TAGS) $(LDFLAGS) $(GCFLAGS) $(PKG_CARBONAPI)
-	$(PKGCONF) $(GO) build $(TAGS) $(LDFLAGS) $(GCFLAGS) $(PKG_CARBONZIPPER)
+	$(PKGCONF) $(GO) build -mod vendor $(TAGS) $(LDFLAGS) $(GCFLAGS) $(PKG_CARBONAPI)
```
avereha, line 37: We already have:
https://github.com/bookingcom/carbonapi/blob/b0a55db0d45f14ac09adcabc85d6e5ddda423775/Makefile#L9
Why do you think it is no working?

go.sum
```diff
@@ -194,7 +194,6 @@ go.opencensus.io v0.21.0/go.mod h1:mSImk1erAIZhrmZN+AvHh14ztQfjbGwt4TtuofqLduU=
 go.opencensus.io v0.22.0/go.mod h1:+kGneAE2xo2IficOXnaByMWTGM9T73dGwxeWcUqIpI8=
 go.opencensus.io v0.22.2/go.mod h1:yxeiOL68Rb0Xd1ddK5vPZ/oVn4vY4Ynel7k9FzqtOIw=
 go.opencensus.io v0.22.3/go.mod h1:yxeiOL68Rb0Xd1ddK5vPZ/oVn4vY4Ynel7k9FzqtOIw=
-go.opentelemetry.io v0.1.0 h1:EANZoRCOP+A3faIlw/iN6YEWoYb1vleZRKm1EvH8T48=
```
avereha, line 197: Would it make sense to put `go mod tidy` in CI. And fail if there are any changes.

grzkv: +1

Makefile
grzkv: Sorry, did not notice that one. But frankly, I would rather do it explicitly in the command. Evidence: I missed it.

grzkv: Pushed some more updates.

avereha: I think that forgo 1.14 we don't even need it(see https://golang.org/doc/go1.14#go-command), but if you prefer using it we can keep it.
Would it make sense to set the same flag also for the `test` target for consistency?

grzkv: Frankly, I'd keep it because this keeps changing every release)

PR #275, opened 2020-08-05
Avereha/error

avereha:

**Logic change**: when we have a render request and one query fails to parse, just answer with http 400 and return that error message. Do not answer with partial results. Before, we had the same behavior only if parsing failed during `ParseExpr`, but when parsing failed during `getTargetData` we would ignore this error. 
I also tested how does graphite-web handle this and it seems to answer with the first http 500 for the parsing error that it finds.

 - Answer with HTTP 400 for parsing errors during render requests. The error message will contain only the first parsing error encountered. Example response: `Bad Request (400) Details: expression eval failed: missing argument "`

 - Cleanup mostly not used pkg/error dependency

 - group all parsing error under ParserError type so we could check later if one error has this type or not.
 - since I started using `errors.As` for this check, I removed `pkg/errors` dependency because it was not used anyway(we would wrap errors sometime, but we would not use that wrapping for anything) and I also wanted to avoid the mix of Cause and As/Is.

I think the logic change that I mentioned earlier is the best approach because that way we have consistent HTTP response for parsing errors.

Tested by hand for now. 
I Will test tomorrow with production traffic(and edit this message when it's done)


app/carbonapi/app.go
```diff
@@ -498,7 +497,7 @@ func initBackend(config cfg.API, logger *zap.Logger) (backend.Backend, error) {
 
 	// TODO (grzkv): Stop using a list, move to a single value in config
 	if len(config.Backends) == 0 {
-		return nil, errors.New("got empty list of backends from config")
+		return nil, fmt.Errorf("got empty list of backends from config")
```
grzkv, line 500: There are no arguments for string, so `errors.New` is sufficient.

app/carbonapi/app_test.go
```diff
@@ -64,7 +64,7 @@ func render(ctx context.Context, request types.RenderRequest) ([]types.Metric, e
 func renderErr(ctx context.Context, request types.RenderRequest) ([]types.Metric, error) {
 	return []types.Metric{
 		{},
-	}, errors.New("error during render")
+	}, fmt.Errorf("error during render")
```
grzkv, line 67: Again, nothing in terms of formatting and arguments. `errors.New` is the simpler alternative.

app/carbonapi/http_handlers_test.go
```diff
@@ -51,7 +51,7 @@ func TestOptimistErrsFanIn(t *testing.T) {
 		{
 			name: "1 err, 1 result",
 			in: []error{
-				errors.New("some error"),
+				fmt.Errorf("some error"),
```
grzkv, line 54: same

app/carbonapi/http_handlers_test.go
```diff
@@ -76,7 +76,7 @@ func TestOptimistErrsFanIn(t *testing.T) {
 		{
 			name: "2 mixed errs, 2 results",
 			in: []error{
-				errors.New("some error"),
+				fmt.Errorf("some error"),
```
grzkv, line 79: same

app/carbonapi/http_handlers_test.go
```diff
@@ -102,7 +102,7 @@ func TestOptimistErrsFanIn(t *testing.T) {
 		{
 			name: "1 arbitrary err, 2 results",
 			in: []error{
-				errors.New("some err"),
+				fmt.Errorf("some err"),
```
grzkv, line 105: same

app/carbonapi/http_handlers_test.go
```diff
@@ -146,7 +146,7 @@ func TestPessimistErrsFanIn(t *testing.T) {
 		},
 		{
 			name:        "one in",
-			in:          []error{errors.New("some error")},
+			in:          []error{fmt.Errorf("some error")},
```
grzkv, line 149: same

app/carbonapi/http_handlers.go
```diff
@@ -214,9 +215,23 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request) {
 		targetErr, metricSize := app.getTargetData(targetCtx, target, exp, metricMap,
 			&results, &form, &toLog, logger, &partiallyFailed)
 		if targetErr != nil {
+			// we can have 3 error types here
+			// a) dataTypes.ErrNotFound  > Continue, at the end we check if all errors are 'not found' and we answer with http 404
+			// b) parser.ParseError -> Return with this error(like above, but with less details )
+			// c) anything else -> continue, answer will be 5xx if all targets have one error
```
grzkv, line 221: I think returning with 5xx makes more sense.

app/carbonapi/app.go
avereha: Done!

app/carbonapi/app_test.go
avereha: Done!

app/carbonapi/http_handlers_test.go
avereha: Done!

app/carbonapi/http_handlers_test.go
avereha: Done!

app/carbonapi/http_handlers.go
avereha: Yes, agree! And then we will not optimistFanIn for target errors. 

I just want to implement this change in smaller steps so we could see better the error types.
After this change, we will see better what type are the remaining errors and what type are they supposed to be. The reason I did not implement it now was that I'm not sure if some of the errors from c) can be 'ErrNotFound' masked as 'generic error' or not.

The plan is to roll this and then we can have better view on remaining errors.

PR #284, opened 2020-08-13
Avereha/errrors 2

avereha:

Addressing this comment: https://github.com/bookingcom/carbonapi/pull/275#discussion_r465954831
When I pushed the previous change, we still had too many errors and it was not clear if we can treat all unknown errors as 'http 500' or not.

Now that we cleaned up almost all errors, it's safer to say that remaining errors are just 'internal server error' 
Since I was editing that part, I also moved all the error reporting details in the 'writeError' function.

make test
tested with production traffic.


grzkv:
Does this PR also include extrapolation besides error logic changes?

grzkv:
Can it be that this PR also includes commits from other PRs? I see a lot of changes that are not errors-related.

avereha:
yes, I have to rebase it.

avereha:
Rebased!

PR #285, opened 2020-08-14
Avereha/golangci lint

avereha:

Add golangci-lint in CI.

Was following docs here: https://golangci-lint.run/usage/install/#ci-installation

Let's see if it does something when I open a PR. It does!
Will continue this after merging other PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,9 @@
-on: [push, pull_request]
 name: CI
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 jobs:
   test:
     strategy:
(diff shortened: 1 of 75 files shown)


grzkv:
We already have linting with the same toolset in our makefile. Wouldn't it be simpler to enable linting in already existing workflow instead of introducing a whole new one?

avereha:
I followed advice from official documentation that is linked. Reasons listed in the docs:
```
We recommend using our GitHub Action for running golangci-lint in CI for GitHub projects. It's fast and uses smart caching inside and it can be much faster than the simple binary installation.
Also, the action creates GitHub annotations for found issues: you don't need to dig into build log to see found by golangci-lint issues:
```

grzkv:
I am not sure the documentation applies in our case or is it for a case when you have only the linting in CI. 

The main upside I can see is speed-up. It is true that this kind of linting can be quite slow. We took care to make it reasonably fast but I frankly don't know the current state. If this workflow runs in parallel and linting takes majority of CI time, it makes sense.

grzkv (approved):

avereha:
Moved it to the same workflow as a different step.
Also configured CI to run only on merge requests and push to master, so we don't have two running tests in parallel.

PR #296, opened 2020-09-09
asPercent: rewrite to support different resolution metrics

avereha:

Rewrite asPercent to support metrics with different resolutions.

Part of this change:
 - update signature of CombineSeries callback to return both (value,
   isAbsent).
 - added constructor for MetricData.
 - updated AggregateSeries signature to received the name of the
 resulting metric. 

Added new tests(imported from original graphite test suite).

### DIFF
--- a/expr/types/types.go
+++ b/expr/types/types.go
@@ -31,6 +31,20 @@ type MetricData struct {
 	AggregateFunction func([]float64, []bool) (float64, bool)
 }
 
+// New creates new MetricData with given metric timeseries. values have math.NaN() for absent
+func New(name string, values []float64, isAbsent []bool, step, start int32) *MetricData {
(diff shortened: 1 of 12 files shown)


expr/types/types.go
```diff
@@ -31,6 +31,20 @@ type MetricData struct {
 	AggregateFunction func([]float64, []bool) (float64, bool)
 }
 
+// New creates new MetricData with given metric timeseries. values have math.NaN() for absent
```
grlvrl, line 34: looks like we don't have  math.NaN() for absent values, but keep a boolean array of absence.

grlvrl (approved):

expr/types/types.go
avereha: Indeed, updated comment!

PR #308, opened 2020-09-18
WriteTimeout & ctx.Deadline race condition

avereha:
#What issue is this change attempting to solve?
When processing a request, we use a ctx with a timeout that was
Timeout.Global value.

At the same time, we set the same Timeout as WriteTimeout when creating
the httpServer.

What happens it that after `Timeout.Global` seconds:
 - we check if ctx.Done() and return from functions with errors.
 - the http server is closing the connection because we reached
   WriteTimeout

So when we try to answer with an error because we reached ctx.Deadline the connection will be closed.

I'm not 100% sure that 2*Timeout.Global is the right value here.
Idea is that WriteTimeout has to be greater than ctx.Deadline, so I though to use the double of that number.

Tested with very slow request that times out.

### DIFF
--- a/app/carbonapi/app.go
+++ b/app/carbonapi/app.go
@@ -113,7 +113,7 @@ func (app *App) Start() func() {
 		Addr:         app.config.Listen,
 		Handler:      handler,
 		ReadTimeout:  1 * time.Second,
-		WriteTimeout: app.config.Timeouts.Global * 2,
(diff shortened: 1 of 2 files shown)


grzkv (approved):

grzkv:
@avereha Please add a comment explaining the logic behind this and maybe explaining the double-cancellation situation that we have.

avereha:
Added comment

PR #317, opened 2020-09-23
Avereha/zipper ctx

avereha:
zipper: 
move handler_test to http_handler_test to match the tested filename
group errors by message and count them


app/carbonzipper/http_handlers.go
```diff
@@ -607,9 +609,10 @@ func errorsFanIn(ctx context.Context, errs []error, nBackends int) error {
 				"majority of backends returned not found. %d total errors, %d not found",
 				nErrs, nErrs-nNotNotFounds))
 		}
-
-		return errors.New(fmt.Sprintf(
-			"all backends failed with mixed errors: %d total errors, %d not found",
-			nErrs, nErrs-nNotNotFounds))
+		message := fmt.Sprintf("all backends failed with mixed errors: %+v", counts)
+		if len(message) > 300 {
```
grlvrl, line 613: Why do we have a hard-coded "300" here?

avereha: Was thinking to avoid huge logs.

grzkv: I'd rather do that on a higher level to have this for all error messages at the same time, but that can be done later.

grzkv (approved):

app/carbonzipper/http_handlers.go
avereha: Filled https://github.com/bookingcom/carbonapi/issues/318 to continue this discussion.

PR #321, opened 2020-09-25
summarize: add support for 'count' and validate aggregation function

avereha:

summarize: add support for 'count' and validate aggregation function

Moved tests from expr_test to this function directory.

Added test for 'count'


expr/helper/helper.go
```diff
@@ -147,11 +147,11 @@ func AggregateSeries(name string, args []*types.MetricData, absent_if_first_seri
 }
 
 // SummarizeValues summarizes values
-func SummarizeValues(f string, values []float64) (float64, bool) {
+func SummarizeValues(f string, values []float64) (error, float64, bool) {
```
grzkv, line 150: By convention, errors should be the last return value. See https://github.com/golang/lint/blob/738671d3881b9731cc63024d5d88cf28db875626/lint.go#L1245

expr/helper/helper.go
```diff
@@ -147,11 +147,11 @@ func AggregateSeries(name string, args []*types.MetricData, absent_if_first_seri
...
 	rv := 0.0
 
 	if len(values) == 0 {
-		return 0, true
+		return nil, 0, true
 	}
 
 	switch f {
```
grzkv, line 157: Looks like, we have a difference vs the python version https://graphite.readthedocs.io/en/latest/functions.html?highlight=summarize#graphite.render.functions.summarize I don't see `total` being a summarization type, and `average` is used instead of `avg`.

expr/helper/helper.go
```diff
@@ -183,18 +183,23 @@ func SummarizeValues(f string, values []float64) (float64, bool) {
...
 		}
+	case "count":
+		rv = float64(len(values))
 
 	default:
-		f = strings.Split(f, "p")[1]
-		percent, err := strconv.ParseFloat(f, 64)
-		if err == nil {
-			return Percentile(values, percent, true)
+		if strings.HasPrefix(f, "p") {
```
grzkv, line 190: Do we really need the percentile summarization type that is not in the [reference version](https://graphite.readthedocs.io/en/latest/functions.html?highlight=summarize#graphite.render.functions.summarize)? Even if we do, it makes sense to add the `median` that is present in reference, which we can get for free.

grzkv, line 190: This check seems partial. "p" being a prefix does not mean we have a correct percentile. In case we get something like "puff", the returning error will be misleading. I suggest we change the check to something like `^p\d\d?$`.

What do you think?

grzkv (changes requested):

expr/helper/helper.go
avereha: Done!

expr/helper/helper.go
avereha: I think the documentation is a bit off.  I was reading the source code, and this is what graphite is using:

https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/functions/aggfuncs.py

`total` is an alias for `sum`.
Added `average` as an alias for `avg`.

Another difference that I noticed is that draphite doesn't seem to support percentiles(see the default: in our case).

expr/helper/helper.go
avereha: I don't think they are very useful, but since we have them they might be used :(.
Added `median`

expr/helper/helper.go
avereha: Also added `current` as alias for `last`.
Filled https://github.com/bookingcom/carbonapi/issues/323 for the missing aggregations.

expr/helper/helper.go
avereha: I like this idea!
I used `^p\d` because it seems we can have float percentiles here. For example: `p99.99'

expr/helper/helper.go
grzkv: @avereha Does "current" mean the same to you as "last"? I'd avoid using "current" if it's not in the docs. It's misleading.

expr/helper/helper.go
```diff
@@ -179,22 +178,33 @@ func SummarizeValues(f string, values []float64) (float64, bool) {
...
+		rv = float64(len(values))
+	case "median":
+		val, absent := Percentile(values, 50, true)
+		return val, absent, nil
 	default:
-		f = strings.Split(f, "p")[1]
-		percent, err := strconv.ParseFloat(f, 64)
-		if err == nil {
-			return Percentile(values, percent, true)
+		looks_like_percentile, err := regexp.MatchString(`^p\d`, f)
```
grzkv, line 191: Did you prefer this instead of the full test because you think we could support float percentiles? Let's only support integer percentiles, but not `p100`. What do you think? Then you can do a full check with `^p\d\d?$`.

This will still give a weird error when given something like "p2m".

grzkv (changes requested):

expr/helper/helper.go
avereha: As I mentioned before, documentation for `summarize` is not up to date.
The functions supported by `summarize` are exactly the same function supported by `aggregate`.
This list of functions is more up to date: https://graphite.readthedocs.io/en/latest/functions.html?highlight=summarize#graphite.render.functions.aggregate
And yes, `current` == `last`.

expr/helper/helper.go
avereha: Yes, I discovered this while running tests, see : https://github.com/bookingcom/carbonapi/blob/avereha/summarize/expr/functions/summarize/function_test.go#L128
Not sure what to say, if we have a test for this maybe somebody is using it.
Between breaking functionality that someone is using and "weird error" in very extreme cases, I would choose the second one.

avereha: In the end, we can search the logs and see if anyone is using fractional percentiles. 
I agree that they don't seem very useful, so I'm going to implement you suggestion.

avereha: Done.

grzkv (approved):

PR #325, opened 2020-10-06
traces: add two new options:  jaegerBufferMaxCount and jaegerBatchMaxCount

avereha:

I noticed that `carbonzipper` was running out of memory and getting OOM killed. 
A big chuck of memory was used by spans.

There are two options that control how many spans can get acumulated in memory:
 `BatchMaxCount`(default 10) : maximum number of spans that get sent in a request.
 `BufferMaxCount`(default 1 billion!) : maximum number of spans that can be stored in memory if sending is low.
Becuase of small `BatchMaxCount` spans would get send slow. Because of bug `BufferMaxCount`, the process would run out memory.
See also comment:
https://github.com/open-telemetry/opentelemetry-go/blob/master/exporters/trace/jaeger/jaeger.go#L136

I added two new options:  jaegerBufferMaxCount and jaegerBatchMax.
That way, we can configure max number of spans stored in memory and number of spans that get sent in a request.
Default values are for a span size of ~3k.

make test. 
Tested with prod traffic.

### DIFF
--- a/cfg/common.go
+++ b/cfg/common.go
@@ -120,6 +120,9 @@ func DefaultCommonConfig() Common {
 		Traces: Traces{
 			Timeout: 10 * time.Second,
 			Tags:    Tags{},
+			JaegerBufferMaxCount: 500000, // If size of one span is 3k, we will hold max ~1.5g in memory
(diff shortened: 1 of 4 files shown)


grlvrl (approved):
LGTM

PR #326, opened 2020-12-10
return 404 instead of 500 when info-ing a non-existing metric

dams:


avereha (approved):

pkg/backend/net/net.go
```diff
@@ -392,6 +392,11 @@ func (b Backend) Info(ctx context.Context, request types.InfoRequest) ([]types.I
 	request.Trace.AddMarshal(t0)
 
 	_, resp, err := b.call(ctx, request.Trace, u, body)
+
+	if code, ok := err.(ErrHTTPCode); ok && code == http.StatusNotFound {
```
avereha, line 396:
```suggestion
	if errors.Is(err, http.StatusNotFound) {
```

PR #327, opened 2020-12-14
use a global variable to check errorNotFound type

dams:
Avoid creating local vars just to test the type of errors


grzkv (changes requested):
I am not sure about this change. For several reasons:

1. Global variables are bad idea. There should be a very strong argument about using them.
2. Why would we increase the scope of something that is only used locally? This only creates confusion.
3. Approach of using local err variables like that is a typical Go strategy and is used in many projects. For good reasons.
4. The `dataTypes` package does not make any sense. This is a reflection of the bad strategy of using globals. It should be dissolved, not added to.

dams:
I agree 👍 
But isn't there any other way than having all these local variables that are useless (we use them only for their type), and duplicated all over the place ?
Maybe delegate that to a function ? but that's not really better imho

grzkv:
@dams Yes it's far from ideal. It's a long standing discussion. This is probably the best write-up to get the grip on the situation https://blog.golang.org/go1.13-errors

PR #340, opened 2021-09-22
carbonzipper: introduce deterministicBackend option

bom-d-van:

In graphite storage (go-carbon) installations where consistency can not be guaranteed,
by always prefer to return metrics from the same set of backends would make the query
result less confusing, no more flapping results.

This commit essentially make graphite render result more consistent, both and right
and wrong.

DeterministicBackend sets a backend weight for each metrics in render
request and carbonzipper would favor metrics from a specific node.
As storage backends like go-carbon does not have strong consistency
guarantee, this means that the same metric persisted in different
replicas might have different values. Without DeterministicBackend
option, carbonzipper would just favor metrics returned first from
any backend. This might result in confusing query result in
carbonapi that are reading multiple metrics in one query.

An example like bellow:

```
asPercent(
    integral(sumSeries(http.app.latencies.{0_5_1,1_5,5_10,10_25,25_50,50_100,100_250}.count)),
    integral(sumSeries(http.app.latencies.*.count))
)
```

Without DeterministicBackend option, the example query above might
return result 100.1%. With DeterministicBackend option, it is
guaranteed that the query result would always "seems" meaningful,
but not necessarily correct.

Check TestMergeMetricsDifferingStepTimes7. Also confirm with production data that returning results are always consistent now.

### DIFF
--- a/cfg/common.go
+++ b/cfg/common.go
@@ -118,10 +118,10 @@ func DefaultCommonConfig() Common {
 			},
 		},
 		Traces: Traces{
-			Timeout: 10 * time.Second,
-			Tags:    Tags{},
+			Timeout:              10 * time.Second,
+			Tags:                 Tags{},
 			JaegerBufferMaxCount: 500000, // If size of one span is 3k, we will hold max ~1.5g in memory
-			JaegerBatchMaxCount: 500, // If size of one span is 3k, total request size will be ~1.5m
+			JaegerBatchMaxCount:  500,    // If size of one span is 3k, total request size will be ~1.5m
 
 		},
 		PrintErrorStackTrace: false,
@@ -167,6 +167,31 @@ type Common struct {
 
 	Traces               Traces `yaml:"traces"`
 	PrintErrorStackTrace bool   `yaml:"printErrorStackTrace"`
+
+	// DeterministicBackend sets a backend weight for each metric in render
+	// request and carbonzipper would favor metrics from a specific node.
+	// As storage backends like go-carbon doesn't have strong consistency
+	// guarantee, this means that the same metric persisted in different
+	// replicas might have different values. Without DeterministicBackend
+	// option, carbonzipper would just favor metrics returned first from
+	// any backend. This might result in confusing query result in
+	// carbonapi that are reading multiple metrics in one query.
+	//
+	// An example like bellow:
+	//
+	//     asPercent(
+	//         integral(sumSeries(http.app.latencies.{0_5_1,1_5,5_10,10_25,25_50,50_100,100_250}.count)),
+	//         integral(sumSeries(http.app.latencies.*.count))
+	//     )
+	//
+	// Without DeterministicBackend option, if there are inconsistent values
+	// between replicas in the returning metrics (for example, in one
+	// replica 0_5_1.count has value 10, and in another 0_5_1.count has 5),
+	// the example query above might return percentage values larger than
+	// 100%. With DeterministicBackend option, it's guaranteed that the
+	// query result would always "seem" meaningful (in the example, it
+	// always return value <= 100%), but not necessarily correct.
+	DeterministicBackend bool `yaml:"deterministicBackend"`
 }
 
 // GetBackends returns the list of backends from common configuration
--- a/pkg/backend/net/net.go
+++ b/pkg/backend/net/net.go
@@ -5,6 +5,7 @@ package net
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -53,6 +54,7 @@ type Backend struct {
 	logger         *zap.Logger
 	cache          *expirecache.Cache
 	cacheExpirySec int32
+	deterministic  bool // Check cfg/common.go:Common.DeterministicBackend
 }
 
 // Config configures an HTTP backend.
@@ -73,6 +75,8 @@ type Config struct {
 	Logger             *zap.Logger   // Logger to use. Defaults to a no-op logger.
 	ActiveRequests     prometheus.Gauge
 	WaitingRequests    prometheus.Gauge
+
+	Deterministic bool // Check cfg/common.go:Common.DeterministicBackend
 }
 
 var fmtProto = []string{"protobuf"}
@@ -98,6 +102,7 @@ func New(cfg Config) (*Backend, error) {
 	b.scheme = scheme
 	b.cluster = cfg.Cluster
 	b.dc = cfg.DC
+	b.deterministic = cfg.Deterministic
 
 	if cfg.Timeout > 0 {
 		b.timeout = cfg.Timeout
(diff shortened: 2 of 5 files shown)


cfg/common.go
```diff
@@ -167,6 +167,28 @@ type Common struct {
 
 	Traces               Traces `yaml:"traces"`
 	PrintErrorStackTrace bool   `yaml:"printErrorStackTrace"`
+
+	// DeterministicBackend sets a backend weight for each metrics in render
```
timtofan, line 171: I think it should be either `each metric` or `all metrics`.

cfg/common.go
```diff
@@ -167,6 +167,28 @@ type Common struct {
...
+	// An example like bellow:
+	//
+	//     asPercent(
+	//         integral(sumSeries(http.app.latencies.{0_5_1,1_5,5_10,10_25,25_50,50_100,100_250}.count)),
+	//         integral(sumSeries(http.app.latencies.*.count))
+	//     )
+	//
+	// Without DeterministicBackend option, the example query above might
+	// return result 100.1%. With DeterministicBackend option, it's
+	// guaranteed that the query result would always "seems" meaningful,
```
timtofan, line 189: result would `seems` vs result would `seem`

pkg/types/types.go
```diff
@@ -231,7 +239,21 @@ func (s byStepTime) Swap(i, j int) {
 }
 
 func (s byStepTime) Less(i, j int) bool {
-	return s[i].StepTime < s[j].StepTime
+	// no deterministic backend sorting
+	if s[i].StepTime != s[j].StepTime || s[i].backendAddr == "" {
+		return s[i].StepTime < s[j].StepTime
+	}
+
+	if s[i].backendWeight < s[j].backendWeight {
```
timtofan, line 247: What's the advantage of comparing metricName+address hashes as opposed to just checking addresses as happens in `// dealing with hash collision`?

pkg/types/types.go
```diff
@@ -231,7 +239,21 @@ func (s byStepTime) Swap(i, j int) {
 }
 
 func (s byStepTime) Less(i, j int) bool {
```
timtofan, line 241: I think we can use sort.Slice instead of special type for sorting. It [was introduced in go 1.8](https://go.dev/doc/go1.8#sort_slice). OTOH maybe it should be a separate patch 🤷

pkg/types/types.go
bom-d-van: the downside of just sorting by server address is that we might potentially always return metrics in a dc/az due to it's hostnames or ip addresses.
I think it's better to uniformly retrieve metrics from different replicas but still stick to the same server in every api calls.

pkg/types/types.go
bom-d-van: yep. could be a separate change. and as a separate function also isn't bad, we can reuse it somewhere else if needed.

cfg/common.go
```diff
@@ -167,6 +167,28 @@ type Common struct {
...
+	//
+	// An example like bellow:
+	//
+	//     asPercent(
+	//         integral(sumSeries(http.app.latencies.{0_5_1,1_5,5_10,10_25,25_50,50_100,100_250}.count)),
+	//         integral(sumSeries(http.app.latencies.*.count))
+	//     )
+	//
+	// Without DeterministicBackend option, the example query above might
+	// return result 100.1%. With DeterministicBackend option, it's
```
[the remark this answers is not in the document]
bom-d-van: Not for this example. Reword the comments to make it clear:

> Without DeterministicBackend option, if there are inconsistent values
> between replicas in the returning metrics (for example, in one
> replica 0_5_1.count has value 10, and in another 0_5_1.count has 5),
> the example query above might return percentage values larger than
> 100%. With DeterministicBackend option, it's guaranteed that the
> query result would always "seem" meaningful (in the example, it
> always return value <= 100%), but not necessarily correct.

cfg/common.go
bom-d-van: thanks, fixed now.

cfg/common.go
bom-d-van: thanks, fixed now.

cfg/common.go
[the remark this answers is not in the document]
timtofan: Hmm, sorry, after rephrase it didn't get any clearer for me 😬

pkg/types/types.go
timtofan: I see. I think given it's not obvious, maybe worth adding a comment.

Would it be sufficient to hash address w/ metric name to achieve the same result?

bom-d-van: > I see. I think given it's not obvious, maybe worth adding a comment.

yep, sounds good.

> Would it be sufficient to hash address w/o the metric name to achieve the same result?

then it's just the same as sorting backend address?

bom-d-van: Adds the comment here: https://github.com/bookingcom/carbonapi/pull/340/files#diff-9b4292c30cc53778b4e267af519527527024f6c158979cb0a0a55ca0192651a0R162-R165

cfg/common.go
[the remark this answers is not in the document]
bom-d-van:
```
asPercent(
    integral(sumSeries(http.app.latencies.{0_5_1,1_5,5_10,10_25,25_50,50_100,100_250}.count)),
    integral(sumSeries(http.app.latencies.*.count))
)
```

Let's suppose there are two replicas and only metric `0_5_1.count` has inconsistent values at `timestamp T` (`10` in replica a, `5` in replica b).

Without `DeterministicBackend`, carbonzipper might fetches `0_5_1.count` for `integral(sumSeries(http.app.latencies.{0_5_1,1_5,5_10,10_25,25_50,50_100,100_250}.count)),` from replica a, and fetches `0_5_1.count` for `integral(sumSeries(http.app.latencies.*.count))` from replica b.

These means that at `timestamp T`, `integral(sumSeries(http.app.latencies.{0_5_1,1_5,5_10,10_25,25_50,50_100,100_250}.count))` is larger than `integral(sumSeries(http.app.latencies.*.count))`, and it would produces a percentage value >= 100%.

With `DeterministicBackend`, carbonzipper would either use replica a or replica b for both subqueries thus its guarantee to return a percentage value <= 100%, which is arguably more meaningful.

grzkv (dismissed):
@bom-d-van If I understand everything correctly, I am not sure this adds value to the system. Consistently returning a wrong result could be considered worse than inconsistently returning results. Inconsistency still is an indicator to the user - even if awkward - that something is not right.

Surely, it is better to let the user know that the data is wrong. This change hides the errors the system makes. It makes bad data look good.

cfg/common.go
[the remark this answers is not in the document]
timtofan: ohh, yep got it now, thx

pkg/types/types.go
timtofan: > then it's just the same as sorting backend address?
true, nvm

pkg/backend/net/net.go
```diff
@@ -154,6 +159,16 @@ func (b Backend) GetServerAddress() string {
 	return b.address
 }
 
+// Motivations of producing a weight by hashing metric_name + server_addr:
+//
+// 	Make carbonzipper uniformly retrieving metrics from different
```
timtofan, line 164: `retrieving` vs `retrieve`

pkg/backend/net/net.go
```diff
@@ -154,6 +159,16 @@ func (b Backend) GetServerAddress() string {
...
+// 	replicas but still stick to the same backend server in every api calls.
```
timtofan, line 165: `every api calls` vs `every api call`

pkg/backend/net/net.go
bom-d-van: 🤦 thanks.

PR #343, opened 2021-12-23
Emadolsky/weighted average

emadolsky:

Currently, carbonapi does not support weightedAverage function.

This PR adds weightedAverage function to carbonapi.

Tests are included. Tests are copied and used from https://github.com/go-graphite/carbonapi/blob/main/expr/functions/weightedAverage/function_test.go.

### DIFF
--- a/expr/functions/weightedAverage/function.go
+++ b/expr/functions/weightedAverage/function.go
@@ -65,11 +65,14 @@ func (f *weightedAverage) Do(ctx context.Context, e parser.Expr, from, until int
 
 	var productMetrics []*types.MetricData
 	for _, key := range keys {
-		for i, _ := range avgGrouped[key] {
-			productMetricName := fmt.Sprintf("product(%s, %s)", weightGrouped[key][i].Name, avgGrouped[key][i].Name)
-			productMetric := helper.CombineSeries(avgGrouped[key][i], weightGrouped[key][i], productMetricName, productOperator)
-			productMetrics = append(productMetrics, productMetric)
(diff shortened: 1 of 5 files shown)


expr/functions/weightedAverage/function.go
```diff
@@ -0,0 +1,163 @@
...
+
+	var productMetrics []*types.MetricData
+	for _, key := range keys {
+		if len(avgGrouped[key]) == 0 || len(weightGrouped[key]) == 0 {
+			continue
+		}
+		weight := weightGrouped[key][len(weightGrouped[key])-1]
+		avg := avgGrouped[key][len(avgGrouped[key])-1]
+		productMetricName := fmt.Sprintf("product(%s, %s)", weight.Name, avg.Name)
+		productMetric := helper.CombineSeries(avg, weight, productMetricName, productOperator)
```
bom-d-van, line 74: is there any specific reason that you are using `helper.CombineSeries` rather than `helper.AggregateSeries` here?

bom-d-van (approved):
lgtm. left a small comment, but not a blocker.

expr/functions/weightedAverage/function.go
emadolsky: well no. Since it was only two timeseries for the operation, I used `helper.CombineSeries`. is there a concern in using it?

bom-d-van: I was reading both functions and it seems `helper.CombineSeries` and `helper.AggregateSeries` are doing similar things and `helper.CombineSeries` doesn't exist in `go-graphite/carbonapi`. Probably a refactoring target.

PR #345, opened 2022-01-11
WIP: consistency report

emadolsky:
There are reports of inconsistencies in graphite metrics.
This commit aims at adding functionality for better tracing of the
inconsistencies on render queries.

It exposes metrics on how much inconsistencies we offer with the current implementation.
It gives us visibility on consistency.
The performance loss is a big deal however. There should be tests for validating that it is worth it. 

Tests for the new logic are added.


grzkv (changes requested):
@emadolsky As we have discussed on our meeting, please make the metrics more universal. We can start simply with a pair of counters: total points and inconsistent points.

Also, you can mark PR as draft if it's WIP. I'll do it for this PR just to illustrate.

app/carbonzipper/http_handlers.go
```diff
@@ -309,7 +309,9 @@ func (app *App) renderHandler(w http.ResponseWriter, req *http.Request) {
 	request.Trace.OutDuration = app.prometheusMetrics.RenderOutDurationExp
 	bs := app.filterBackendByTopLevelDomain(request.Targets)
 	bs = backend.Filter(bs, request.Targets)
-	metrics, errs := backend.Renders(ctx, bs, request)
+	metrics, points, incons, errs := backend.Renders(ctx, bs, request, app.config.ConsistencyCheck)
+	app.prometheusMetrics.RenderConsistency.WithLabelValues("all").Add(float64(points))
```
grzkv, line 313: Could we use int instead of `float64`?

app/carbonzipper/metrics.go
```diff
@@ -85,6 +86,13 @@ func NewPrometheusMetrics(config cfg.Zipper) *PrometheusMetrics {
 			},
 			[]string{"code", "handler"},
 		),
+		RenderConsistency: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "render_consistency",
+				Help: "Count of rendered data points, partitioned by the consistency state",
```
grzkv, line 92: Please add label descriptions. This will make it clear how to use it.

pkg/types/types.go
```diff
@@ -197,13 +197,18 @@ type Metric struct {
...
-		return metrics[0]
+		pointCount := 0
+		ms := metrics[0]
+		for _, m := range ms {
+			pointCount += len(m.Values)
+		}
+		return ms, pointCount, 0
 	}
 
 	names := make(map[string][]Metric)
```
grzkv, line 214: Could you please refactor this name to reflect what variable does please?

pkg/types/types.go
```diff
@@ -234,39 +244,75 @@ func (s byStepTime) Less(i, j int) bool {
...
 				healed++
-				break
+				pointExists = true
+			}
+		}
+		if consistencyCheck {
+			c := getPointInconsistentValueCount(metric.Values[i], valuesForPoint)
+			if c > 0 {
+				inconsistencies++
+				corruptionLogger.Warn("metric point consistency issue",
```
grzkv, line 310: 1. Please consider making logging per-request or per-metrics
2. If logging is potentially high-freq, please check if it's rate-limited

grzkv (changes requested):

pkg/types/types.go
```diff
@@ -234,52 +244,91 @@ func (s byStepTime) Less(i, j int) bool {
...
+		if consistencyCheck {
+			dataPointInconsistentValues := getPointInconsistentValueCount(metric.Values[i], valuesForPoint)
+			if dataPointInconsistentValues > 0 {
+				inconsistencies++
 			}
 		}
 	}
 
+	if consistencyCheck && inconsistencies > 0 {
+		corruptionLogger.Warn("metric consistency issue",
```
grzkv, line 315: This can potentially write a log of logs. Two options to solve this:
1. Put rate-limiting on the logger.
2. Log per-request and include a truncated list of inconsistent metrics in the log.

grzkv (changes requested):

pkg/types/types.go
emadolsky: chose 2.

pkg/types/types.go
```diff
@@ -234,39 +262,69 @@ func (s byStepTime) Less(i, j int) bool {
...
+func getPointInconsistentValueCount(selectedValue float64, values []float64) int {
+	valuesCount := len(values)
+	if valuesCount == 0 {
+		// No value for the point
+		return 0
+	}
+
+	valuesToCount := make(map[float64]int)
+	for _, val := range values {
+		valuesToCount[val]++
```
timtofan, line 274: Would counting only selectedValue also work?
E.g.
```
if val == selectedValue {
  count++
fi
```

pkg/types/types.go
```diff
@@ -234,39 +262,69 @@ func (s byStepTime) Less(i, j int) bool {
...
+	}
+
+	return valuesCount - valuesToCount[selectedValue]
```
timtofan, line 277: It seems return value is not important. Perhaps we can return boolean instead?

pkg/types/types.go
```diff
@@ -234,39 +262,69 @@ func (s byStepTime) Less(i, j int) bool {
 	return s[i].StepTime < s[j].StepTime
 }
 
-func mergeMetrics(metrics []Metric) Metric {
+func getPointInconsistentValueCount(selectedValue float64, values []float64) int {
```
timtofan, line 265: It's not very clear what InconsistentValue means w/o reading function implementation. A potentially more obvious naming would be something like `areValuesEqual()`.

pkg/types/types.go
```diff
@@ -234,39 +262,69 @@ func (s byStepTime) Less(i, j int) bool {
...
 	sort.Sort(byStepTime(metrics))
 	healed := 0
 
 	// metrics[0] has the highest resolution of metrics
-	metric := metrics[0]
+	metric = metrics[0]
 	for i := range metric.Values {
-		if !metric.IsAbsent[i] {
-			continue
+		var valuesForPoint []float64
```
timtofan, line 296: What if we just compare each replica to the original value instead of keeping all values in an array? It seems this might simplify the logic somewhat.

pkg/types/types.go
```diff
@@ -197,29 +197,57 @@ type Metric struct {
...
+			metricByNames[m.Name] = append(metricByNames[m.Name], m)
 		}
 	}
 
 	merged := make([]Metric, 0)
-	for _, ms := range names {
-		merged = append(merged, mergeMetrics(ms))
+	pointCount := 0
+	inconsistencyCount := 0
+	InconsistentMetricsLogLimit := 10
```
timtofan, line 225: As you know, capital 1st carries a special in go. Is it required here?

pkg/types/types.go
```diff
@@ -197,29 +197,57 @@ type Metric struct {
...
+	var limitedInconsistentMetricsForLog []map[string]interface{}
```
timtofan, line 226: nit: idiomatic go gravitates toward shorter variable names, especially when it's used only a few lines around declaration.

pkg/types/types.go
emadolsky: No. joust a human error :D

pkg/types/types.go
```diff
@@ -197,29 +197,57 @@ type Metric struct {
 }
 
 // MergeMetrics merges metrics by name.
```
timtofan, line 199: I think a word or two about what `consistencyCheck` arg means would help a lot to understand its meaning w/o having to read the full implementation.

pkg/types/types.go
emadolsky: Well, because I had kept an eye on the future implementation for handling inconsistencies (majority read, for instance), keeping the values and processing them later might be a good idea.

pkg/types/types.go
emadolsky: Yeah. it's all because at first, the function was doing a bit more. right now, it totally makes sense.

pkg/types/types.go
emadolsky: Yeah. it's all because at first, the function was doing a bit more. right now, it totally makes sense.

pkg/types/types.go
emadolsky: Legit.

pkg/types/types.go
emadolsky: hmmm. I always prefer descriptive names for the variables. It helps to know what's the purpose of that thing in advance. But I see your point. I'll try to beautify it at least :))

pkg/types/types.go
emadolsky: Absolutely.

pkg/types/types.go
```diff
@@ -197,29 +197,57 @@ type Metric struct {
...
+		}
+		merged = append(merged, m)
+		inconsistencyCount += c
+		pointCount += len(m.Values)
 	}
 
-	return merged
+	if consistencyCheck && inconsistencyCount > 0 {
+		corruptionLogger.Warn("metric consistency issue",
+			zap.Any("inconsistentMetric", limitedInconsistentMetricsForLog),
```
timtofan, line 245: Inconsistency seems to generic a term to understand what's the actual issue. Maybe something like replicaMismatch would be more clear? Other ideas?

emadolsky: Well I agree with you that it is a general term but if you look at it in the context of `Render` and `Merge`, I guess it does the job. But if you don't feel the same, we can consider changing it to something like `replicaMismatch` as you proposed.

pkg/types/types.go
```diff
@@ -197,29 +197,59 @@ type Metric struct {
...
+			metricByNames[m.Name] = append(metricByNames[m.Name], m)
 		}
 	}
 
 	merged := make([]Metric, 0)
-	for _, ms := range names {
-		merged = append(merged, mergeMetrics(ms))
+	pointCount := 0
+	inconsistencyCount := 0
+	inconsMetricReportLimit := 10
```
timtofan, line 227: nit: imho this var is not really necessary, as the meaning of hardcoded value would be quite clear as well

pkg/types/types.go
```diff
@@ -197,29 +197,59 @@ type Metric struct {
...
-			names[m.Name] = append(names[m.Name], m)
+			metricByNames[m.Name] = append(metricByNames[m.Name], m)
 		}
 	}
 
 	merged := make([]Metric, 0)
-	for _, ms := range names {
-		merged = append(merged, mergeMetrics(ms))
+	pointCount := 0
+	inconsistencyCount := 0
```
timtofan, line 226: if the rest of the code shortens to `incons` then we could shorten it here as well for consistency

pkg/types/types.go
timtofan: My vote would be to not add complexity just in case it might be needed in the future. We could always add this later, once the actual need arises.

pkg/types/types.go
```diff
@@ -197,29 +197,59 @@ type Metric struct {
...
 
-	return merged
+	if consistencyCheck && inconsistencyCount > 0 {
+		corruptionLogger.Warn("metric consistency issue",
+			zap.Any("inconsistentMetric", inconsMetricReport),
+			zap.Int("totalInconsistencies", inconsistencyCount),
+		)
+	}
+
+	return merged, pointCount, inconsistencyCount
```
timtofan, line 252: I wonder if `pointCount` and `inconsistencyCount` should be calculated in this func and passed back to cmd/app.

An alternative approach would be to add inconsistency/replicaMismatch bool into struct Metric, assign it inside MergeMetrics, and have cmd/app calculate pointCount and inconsistencyCount by iterating over Metric structs.

@grzkv @bom-d-van wdyt?

pkg/types/types.go
```diff
@@ -197,29 +197,59 @@ type Metric struct {
...
 		}
 	}
 
 	merged := make([]Metric, 0)
-	for _, ms := range names {
-		merged = append(merged, mergeMetrics(ms))
+	pointCount := 0
+	inconsistencyCount := 0
+	inconsMetricReportLimit := 10
+	var inconsMetricReport []map[string]interface{}
```
grzkv, line 228: Sorry @emadolsky but `interface{}` is not ok here as it is not in most places) Please replace with a struct if necessary.
Also, isn't `metricName` unique? If yes, we can simply use a `map[string]struct`.
Also, aren't `start` and `stop` the same for all metrics? If yes, we can put it on top level.

If answers are yes to last 2 questions, we can simply go with `map[string]int` as `step` is not really important.

pkg/types/types.go
grzkv: It is necessary to limit the volume of the logs. But I agree that hard-coding is not great here. Please move to config.

grzkv (changes requested):

pkg/types/types.go
emadolsky: > Also, isn't metricName unique? If yes, we can simply use a map[string]struct.

I think it is more suitable to have predefined keys and differing values in our log schema. This is the main reason that I did not use metricName as the key.

> Also, aren't start and stop the same for all metrics? If yes, we can put it on top level.

I was not sure that they would be the same for all the metrics returned in a render request. So if you verify that it is always the case, you are absolutely right.

emadolsky: > but interface{} is not ok here as it is not in most places

I wasn't quite happy about using it too. But unfortunately I observed that `zap.Any` doesn't work as expected on structs. I can convert all the values to strings, but I prefer not to do that.

pkg/types/types.go
emadolsky: I think moving it to config is an overkill as we probably will not configure it at any points in the future.
I will go for a constant in the same file. Any thoughts?

timtofan: I agree that it doesn't seem like something we would need to configure. Also we could it in the future, if it will be needed. IMHO we don't even need the constant, but could just inline the value, as the meaning would be pretty clear from the code.

@bom-d-van wdyt?

pkg/types/types.go
timtofan: @bom-d-van wdyt?

pkg/types/types.go
bom-d-van: I'm ok with a named constant. actually it's even better to declare it as `const` if it's truly a constant.

```
const inconsMetricReportLimit = 10
```

pkg/types/types.go
bom-d-van: moving them into `struct Metric` does make the function interface cleaner. but I guess the aggregation and logging logics still need to live somewhere.

maybe in the future we can return the summary result as a struct if the interface continue to grow.

timtofan: So which approach do you think would be preferable? The current one or the one I described above?

pkg/types/types.go
bom-d-van: `replicaMismatch` is indeed more descriptive. for some reason, we just keep using inconsistent/inconsistency for the whole issue. if we changed it here, we might have to change all other names with `inconsistent` to `replicaMismatch` in the PR? If so, maybe it's easier for us to stick with `inconsistent`?

pkg/types/types.go
bom-d-van: I think the `struct Metric` way would be nicer. Then we can have a separate function for logging and summary and requires no interface changes for the MergeMetrics function. But I will leave it for Emad to make the call.

pkg/types/types.go
timtofan: If we agree that replicaMismatch is better, it's best to change it once rather than keep less descriptive wording forever.

pkg/backend/benchmark_test.go
```diff
@@ -119,22 +127,28 @@ func BenchmarkRendersStorm(b *testing.B) {
...
-				errs <- err
-			}()
-		}
-
-		wg.Wait()
-		for j := 0; j < n; j++ {
-			err := <-errs
-			if len(err) != 0 {
-				b.Fatal(err)
+	consistencyChecks := []bool{false, true}
```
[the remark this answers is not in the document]
emadolsky: Yes and there is a significant difference (which is about 3 to 4 times slower on storm bench) but it does not necessarily hurt the real result because it is not quite considerable compared to latency of fetching metrics from backends.

pkg/types/types.go
emadolsky: I totally agree that moving the mismatch counts to metric is cleaner. However, I kinda have the feeling that having `mismatch` on `metric` can be conceptually irrational because a single metric data points cannot have mismatch (it cannot be defined for a single metric data point) but data points of a group of metrics merged into one can. So I think maybe encapsulating Metric in a MergedMetric or something called like this would be suitable for what we are discussing here. Any thoughts?
But as we talked with @grzkv before, I think keeping it as simple as possible is benefitial.

pkg/types/types.go
```diff
@@ -17,6 +17,10 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	inconsMetricReportLimit = 10
```
grzkv, line 21: Pls move this to config.

pkg/types/types.go
emadolsky: please check @timtofan.

pkg/types/types.go
grzkv: It's up to you how you do it, but `interface{}` is a red flag unless there's no other choice (like parsing schema-less data, etc.) Please replace it with something else. Again, it's not critical with what specifically you replace it with as long as there's no `interface{}`. I can give you more detail in a conversation.

pkg/types/types.go
emadolsky: please check @timtofan.

pkg/types/types.go
emadolsky: please check @timtofan.

grzkv (changes requested):

pkg/types/types.go
emadolsky: used struct with json tags. lgty?

bom-d-van (dismissed):

app/carbonzipper/metrics.go
emadolsky: I consider this as resolved.

pkg/types/types.go
```diff
@@ -234,40 +267,49 @@ func (s byStepTime) Less(i, j int) bool {
...
 	for i := range metric.Values {
-		if !metric.IsAbsent[i] {
-			continue
-		}
-
-		// found a missing value, look for a replacement
+		pointExists := !metric.IsAbsent[i]
+		areValuesEqual := true
 		for j := 1; j < len(metrics); j++ {
+			if pointExists && !mismatchCheck {
```
timtofan, line 296: I think we can short-circuit here if `areValuesEqual == false` while in `mismatchCheck` mode
```suggestion
			if pointExists && !(mismatchCheck && areValuesEqual) {
```

pkg/types/types.go
```diff
@@ -234,40 +267,49 @@ func (s byStepTime) Less(i, j int) bool {
...
+
+			if !pointExists {
 				metric.IsAbsent[i] = m.IsAbsent[i]
 				metric.Values[i] = m.Values[i]
 				healed++
-				break
+				pointExists = true
 			}
+
+			areValuesEqual = areValuesEqual && metric.Values[i] == m.Values[i]
```
timtofan, line 316: If short-circuiting is added in the beginning of the loop, then `areValuesEqual` check could be dropped here

pkg/types/types.go
```diff
@@ -197,29 +201,58 @@ type Metric struct {
...
+		}
+		merged = append(merged, m)
+		mismatchCount += c
+		pointCount += len(m.Values)
 	}
 
-	return merged
+	if mismatchCheck && mismatchCount > 0 {
+		corruptionLogger.Warn("metric mismatch observed",
+			zap.Any("mismatched_metrics", mismatchedMetricReport),
```
timtofan: `replicaMismatch` is kinda clear w/o context. Simply `mismatch` is not much better than `inconsistency` imho.

pkg/types/types.go
emadolsky: Again, it was something left out with future implementations in mind (i.e. majority voting) to see if we can bear the performance penalty or not. Anyways, will change it.

PR #348, opened 2022-02-15
Add render majority read feature

emadolsky:

It is possible that metrics in different backends have different values for points (replica mismatches). These are mainly caused by aggregating metrics with null points. The null points on backends are also caused by backends being temporarily unavailable. 

`renderReplicaMatchMode` is introduced to configure carbonzipper on how it should behave on merging metrics retrieved from backends.
With `majority` implementation, for each of the data points in a render request result, the values stored in the majority of backends will be returned.
If there is no majority in the values, this code returns ~one of the most voted values~ the first value like normal mode (To be discussed).

Also, one counter is added for monitoring number of fixed mismatches. it is increased whenever a data-point has a majority value in `majority` mode.

Possible next steps:
- Configuration on what to do when there is no majority (either null or different).
- take the number of replicas for each metric into account.
  - Pro: eliminate further checks if the majority condition is already met.
  - Con: we should find a way to inject the data to zipper
- take the exact replicas of a metric into account.
  -  Pro: get rid of old or not related backends for a metric. only check the related backends.
  -  Con: implementation can be complex.

Tests are developed for each of the replica match configurations.

### DIFF
--- a/app/carbonzipper/http_handlers.go
+++ b/app/carbonzipper/http_handlers.go
@@ -309,10 +309,11 @@ func (app *App) renderHandler(w http.ResponseWriter, req *http.Request) {
 	request.Trace.OutDuration = app.prometheusMetrics.RenderOutDurationExp
 	bs := app.filterBackendByTopLevelDomain(request.Targets)
 	bs = backend.Filter(bs, request.Targets)
-	metrics, points, mismatches, errs := backend.Renders(ctx, bs, request,
+	metrics, points, mismatches, fixedMismatches, errs := backend.Renders(ctx, bs, request,
 		app.config.RenderReplicaMatchMode, app.config.RenderMismatchMetricReportLimit)
 	app.prometheusMetrics.Renders.Add(float64(points))
 	app.prometheusMetrics.RenderMismatches.Add(float64(mismatches))
+	app.prometheusMetrics.RenderFixedMismatches.Add(float64(fixedMismatches))
 	err = errorsFanIn(errs, len(bs))
 	span.SetAttribute("graphite.metrics", len(metrics))
 	// time in queue is converted to ms
--- a/app/carbonzipper/http_handlers.go
+++ b/app/carbonzipper/http_handlers.go
@@ -310,7 +310,7 @@ func (app *App) renderHandler(w http.ResponseWriter, req *http.Request) {
 	bs := app.filterBackendByTopLevelDomain(request.Targets)
 	bs = backend.Filter(bs, request.Targets)
 	metrics, points, mismatches, errs := backend.Renders(ctx, bs, request,
-		app.config.RenderMismatchCheck, app.config.RenderMismatchMetricReportLimit)
+		app.config.RenderReplicaMatchMode, app.config.RenderMismatchMetricReportLimit)
 	app.prometheusMetrics.Renders.Add(float64(points))
 	app.prometheusMetrics.RenderMismatches.Add(float64(mismatches))
 	err = errorsFanIn(errs, len(bs))
--- a/app/carbonzipper/http_handlers.go
+++ b/app/carbonzipper/http_handlers.go
@@ -309,11 +309,11 @@ func (app *App) renderHandler(w http.ResponseWriter, req *http.Request) {
 	request.Trace.OutDuration = app.prometheusMetrics.RenderOutDurationExp
 	bs := app.filterBackendByTopLevelDomain(request.Targets)
 	bs = backend.Filter(bs, request.Targets)
-	metrics, points, mismatches, fixedMismatches, errs := backend.Renders(ctx, bs, request,
-		app.config.RenderReplicaMatchMode, app.config.RenderMismatchMetricReportLimit)
-	app.prometheusMetrics.Renders.Add(float64(points))
-	app.prometheusMetrics.RenderMismatches.Add(float64(mismatches))
-	app.prometheusMetrics.RenderFixedMismatches.Add(float64(fixedMismatches))
+	metrics, stats, errs := backend.Renders(ctx, bs, request, app.config.RenderReplicaMatchMode,
+		app.config.RenderMismatchMetricReportLimit)
+	app.prometheusMetrics.Renders.Add(float64(stats.DataPointCount))
+	app.prometheusMetrics.RenderMismatches.Add(float64(stats.MismatchCount))
+	app.prometheusMetrics.RenderFixedMismatches.Add(float64(stats.FixedMismatchCount))
 	err = errorsFanIn(errs, len(bs))
 	span.SetAttribute("graphite.metrics", len(metrics))
 	// time in queue is converted to ms
--- a/app/carbonzipper/http_handlers.go
+++ b/app/carbonzipper/http_handlers.go
@@ -310,7 +310,7 @@ func (app *App) renderHandler(w http.ResponseWriter, req *http.Request) {
 	bs := app.filterBackendByTopLevelDomain(request.Targets)
 	bs = backend.Filter(bs, request.Targets)
 	metrics, stats, errs := backend.Renders(ctx, bs, request, app.config.RenderReplicaMatchMode,
-		app.config.RenderMismatchMetricReportLimit)
+		app.config.RenderReplicaMismatchReportLimit)
 	app.prometheusMetrics.Renders.Add(float64(stats.DataPointCount))
 	app.prometheusMetrics.RenderMismatches.Add(float64(stats.MismatchCount))
 	app.prometheusMetrics.RenderFixedMismatches.Add(float64(stats.FixedMismatchCount))
--- a/cfg/common.go
+++ b/cfg/common.go
@@ -126,8 +126,8 @@ func DefaultCommonConfig() Common {
 		},
 		PrintErrorStackTrace: false,
 
-		RenderReplicaMatchMode:          ReplicaMatchModeNormal,
-		RenderMismatchMetricReportLimit: 10,
+		RenderReplicaMatchMode:           ReplicaMatchModeNormal,
+		RenderReplicaMismatchReportLimit: 10,
 	}
 }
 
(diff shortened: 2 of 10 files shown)


pkg/types/types.go
```diff
@@ -270,7 +271,25 @@ func (s byStepTime) Less(i, j int) bool {
...
 }
 
-func mergeMetrics(metrics []Metric, mismatchCheck bool) (metric Metric, mismatches int) {
+func getPointMajorityValue(values []float64) (majorityValue float64, mismatch bool) {
+	valuesCount := len(values)
+	if valuesCount == 0 {
+		return 0, false
+	}
+
+	valuesToCount := make(map[float64]int)
```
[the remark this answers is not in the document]
emadolsky: I have changed the logic so that this function is called when there is an observed mismatch. Additionally, I've set the initial bucket size of the map.
Also, I ran some tests (benchs) on random data with different solutions (the one proposed by @timtofan, map, sort-iterate) and they are not much different.

pkg/types/types.go
```diff
@@ -270,14 +275,32 @@ func (s byStepTime) Less(i, j int) bool {
 	return s[i].StepTime < s[j].StepTime
 }
 
-func mergeMetrics(metrics []Metric, mismatchCheck bool) (metric Metric, mismatches int) {
+func getPointMajorityValue(values []float64) (majorityValue float64, majorityCount int) {
+	valuesCount := len(values)
+	if valuesCount == 0 {
+		return 0, 0
```
grzkv, line 281: This does not seem like a correct behaviour. No values does not give a zero value.

pkg/types/types.go
```diff
@@ -270,14 +275,32 @@ func (s byStepTime) Less(i, j int) bool {
...
+	}
+
+	valuesToCount := make(map[float64]int, valuesCount)
+	majorityValue = values[0]
+	for _, val := range values {
+		valuesToCount[val]++
```
grzkv, line 287: I assume this does simple floats comparison. It is unsafe to do byte-by-byte floats comparison as the representation may differ for the same number. Check this out https://github.com/golang/go/issues/966 for example.

grzkv (changes requested):

cfg/common.go
```diff
@@ -171,8 +171,13 @@ type Common struct {
...
 	PrintErrorStackTrace bool   `yaml:"printErrorStackTrace"`
 
-	RenderMismatchCheck             bool `yaml:"renderMismatchCheck"`
-	RenderMismatchMetricReportLimit int  `yaml:"renderMismatchMetricReportLimit"`
+	// RenderReplicaMatchMode indicates how carbonzipper merges the metrics from replica backends.
+	// Possible values are `normal`(default), `check`, and `majority`
+	// `normal` ignores the mismatches and only heals null points.
+	// `check` looks for mismatches, and exposes metrics.
+	// `majority` chooses the values of majority of backends in addition to exposing metrics.
+	RenderReplicaMatchMode          ReplicaMatchMode `yaml:"renderReplicaMatchMode"`
```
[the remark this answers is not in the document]
emadolsky: I have a different view on this.
First, exposing mismatches requires change in logic (it should look at values even though a value is found). So, separating value check and majority into this two categories is not quite right. I understand your stance, but I looked at this config as `behavior of zipper on matching data points gathered from replicas`.
Second, It doesn't make sense to enable majority voting and disable mismatch metrics. If you're using majority voting, you are doing enough to simply be able to expose metrics with no extra works done.

pkg/types/types.go
emadolsky: It is never called while a mismatch isn't seen. Anyways, as this is also returning `majorityCount`, It should be obvious for the caller not to use the majorityValue while no supporting value is there (zero count). WDYT?

pkg/types/types.go
```diff
@@ -270,14 +275,32 @@ func (s byStepTime) Less(i, j int) bool {
...
 }
 
-func mergeMetrics(metrics []Metric, mismatchCheck bool) (metric Metric, mismatches int) {
+func getPointMajorityValue(values []float64) (majorityValue float64, majorityCount int) {
+	valuesCount := len(values)
+	if valuesCount == 0 {
+		return 0, 0
+	}
+
+	valuesToCount := make(map[float64]int, valuesCount)
```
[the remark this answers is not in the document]
emadolsky: Maybe my understanding on map cap is not right but we have `valuesCount` of keys at most. So, having this number of buckets is enough, isn't it?
As @timtofan mentioned, it is the hottest path of render requests. This was an easy optimization so I ran a quick benchmark, and it slightly helped. I used it then.

pkg/types/types.go
emadolsky: Nice catch. I'll change the implementation.

pkg/types/types.go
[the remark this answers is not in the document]
grzkv: At most, yes.

pkg/types/types.go
```diff
@@ -270,14 +276,48 @@ func (s byStepTime) Less(i, j int) bool {
...
+	})
+
+	majorityCount = 1
+	majorityValue = values[0]
+
+	currentCount := 1
+	currentValue := values[0]
+	for i := 1; i < len(values); i++ {
+		v := values[i]
+		if math.Abs(v-currentValue) < epsilon {
```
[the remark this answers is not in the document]
emadolsky: I'm not sure really. We can make epsilon a function of how big our values are. It's going to perform better.
I've found a pretty good explanation here:
https://stackoverflow.com/a/32334103
And I think it is more accurate so we might want to use this. wdyt?

cfg/common.go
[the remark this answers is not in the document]
bom-d-van: Seems mostly a disagreement of taste/preference and the end behavior of the program would remain largely the same. The docs/comments has also indicated the differences. I would respect Emad's choice here.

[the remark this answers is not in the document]
timtofan: From my side, I agree w/ Emad that original normal/check/majority option makes it clearer that all those options change a single code path, and, importantly, introduce overhead. I think having two bools would make this less obvious.

app/carbonzipper/http_handlers.go
```diff
@@ -309,10 +309,11 @@ func (app *App) renderHandler(w http.ResponseWriter, req *http.Request) {
 	request.Trace.OutDuration = app.prometheusMetrics.RenderOutDurationExp
 	bs := app.filterBackendByTopLevelDomain(request.Targets)
 	bs = backend.Filter(bs, request.Targets)
-	metrics, points, mismatches, errs := backend.Renders(ctx, bs, request,
-		app.config.RenderMismatchCheck, app.config.RenderMismatchMetricReportLimit)
+	metrics, points, mismatches, fixedMismatches, errs := backend.Renders(ctx, bs, request,
```
timtofan, line 312: This has grown from 2 return value to 5 return values within a span of a few weeks. I wonder if there is place for some kind of struct for this data 🤔

pkg/types/types_test.go
```diff
@@ -215,7 +232,59 @@ func TestMergeManyMinorityMismatchedMetrics(t *testing.T) {
...
+		t.Errorf("Expected 1 mismatched point, got %d", ins)
+	}
+
+	if fixed != 0 {
+		t.Errorf("Expected 0 fixed mismatch point , got %d", fixed)
+	}
+}
+
+func TestMergeManyMinorityMismatchedMetricsWithMajority(t *testing.T) {
+	input := [][]Metric{
```
timtofan, line 257: some of this `input` can be deduplicated amongst the [new] tests?

app/carbonzipper/http_handlers.go
emadolsky: Yeah I totally agree. It's become ridiculous. I would go with something like RenderStats struct. Last time, team told me not to use struct and just return two values. So if you verify that this is a good idea, I'll make the change.

pkg/types/types_test.go
emadolsky: There are a lot of things to improve in tests. It's very basic and naive at the moment. I didn't spend time on making it better because I thought dedicating a time to improve the whole testing in the project makes more sense. wdyt?

app/carbonzipper/http_handlers.go
timtofan: my vote'd be for struct

pkg/types/types_test.go
timtofan: ack

app/carbonzipper/http_handlers.go
emadolsky: Made the change.

pkg/types/types.go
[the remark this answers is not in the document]
emadolsky: Made the change.

pkg/types/types.go
emadolsky: Made the change.

cfg/common.go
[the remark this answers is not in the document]
emadolsky: @grzkv Do you have any huge concerns if we keep it this way?

cfg/common.go
```diff
@@ -171,8 +171,13 @@ type Common struct {
 	Traces               Traces `yaml:"traces"`
 	PrintErrorStackTrace bool   `yaml:"printErrorStackTrace"`
 
-	RenderMismatchCheck             bool `yaml:"renderMismatchCheck"`
-	RenderMismatchMetricReportLimit int  `yaml:"renderMismatchMetricReportLimit"`
+	// RenderReplicaMatchMode indicates how carbonzipper merges the metrics from replica backends.
+	// Possible values are `normal`(default), `check`, and `majority`
+	// `normal` ignores the mismatches and only heals null points.
```
timtofan, line 176: The could be indented to mimick bullet points ([not properly support yet](https://github.com/golang/go/issues/7873)).
```// RenderReplicaMatchMode indicates how carbonzipper merges the metrics from replica backends.
	// Possible values are:
	//
	// * `normal` - ignore the mismatches and only heals null points (default)
	//
	// * `check` - look for mismatches, and exposes metrics
	//
	// * `majority` - choose the values of majority of backends in addition to exposing metrics
```

cfg/common.go
```diff
@@ -126,7 +126,7 @@ func DefaultCommonConfig() Common {
 		},
 		PrintErrorStackTrace: false,
 
-		RenderMismatchCheck:             false,
+		RenderReplicaMatchMode:          ReplicaMatchModeNormal,
```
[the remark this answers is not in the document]
emadolsky: Nope. `normal` is the original old behavior.

app/carbonzipper/http_handlers.go
emadolsky: Can you resolve this if you're ok with this now?

bom-d-van (approved):

pkg/types/types.go
```diff
@@ -270,24 +284,84 @@ func (s byStepTime) Less(i, j int) bool {
...
+			div = absA + absB
+		}
+		return diff/div < epsilon
+	}
+}
+
+func getPointMajorityValue(values []float64) (majorityValue float64, majorityCount int, err error) {
+	valuesCount := len(values)
+	if valuesCount == 0 {
+		return math.NaN(), 0, fmt.Errorf("no value for majority voting")
```
grzkv, line 311: It is very strange to use `NaN` in case of incorrect request. `NaN` maybe not be processed correctly all the time. It is really special. I think using a default type value is preferable.

grzkv, line 311: `fmt.Errorf` can be replaced by `errors.New`. The printf functionality of the function is unused here.

cfg/common.go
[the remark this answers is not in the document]
grzkv: Let's keep it. It is not major.

cfg/common.go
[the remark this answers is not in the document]
timtofan: ah, ack

pkg/types/types.go
emadolsky: So the thing is that we are returning an error at the same time. so it doesn't matter much what we are returning for the values. I understand your concern, but I think we can go ahead.

emadolsky: Anyways, decided to make the change :D

bom-d-van (approved):

grzkv (approved):

PR #351, opened 2022-03-07
Change mismatch logs to contain unfixed ones

emadolsky:

Better logs for mismatches; only log mismatches visible to users.

By changing the logging from mismatches to unfixed mismatches.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -245,13 +245,14 @@ func MergeMetrics(metrics [][]Metric, replicaMatchMode cfg.ReplicaMatchMode, rep
 	var mismatchedMetricReports []metricReport
 	for _, ms := range metricByNames {
 		m, stats := mergeMetrics(ms, replicaMatchMode)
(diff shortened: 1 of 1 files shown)


pkg/types/types.go
```diff
@@ -260,8 +261,9 @@ func MergeMetrics(metrics [][]Metric, replicaMatchMode cfg.ReplicaMatchMode, rep
 		metricsStat.DataPointCount += stats.DataPointCount
 	}
 
-	if metricsStat.MismatchCount > 0 {
-		corruptionLogger.Warn("metric replica mismatch observed",
+	metricsUnfixedMismatchCount := metricsStat.MismatchCount - metricsStat.FixedMismatchCount
+	if metricsUnfixedMismatchCount > 0 {
+		corruptionLogger.Warn("metric unfixed replica mismatch observed",
```
grzkv, line 266: @emadolsky Would you please consider refactoring-out the `corruptionLogger` in the future PRs?

grzkv (approved):

pkg/types/types.go
emadolsky: Yeah sure.

PR #353, opened 2022-03-14
Optimize zipper majority vote

emadolsky:

Regarding the performance issues seen on majority read feature introduced in #348 on carbonzipper rendered replica metrics, this PR attempts to optimize the computation on merge functionality of zipper.

Using the [Boyer–Moore majority vote algorithm](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm) (proposed by @bom-d-van) The performance of majority voting is increased. The benchmark results (with new data) is presented:

before Boyer–Moore majority vote algorithm:
```
goos: linux
goarch: amd64
pkg: github.com/bookingcom/carbonapi/pkg/backend
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-normalWithExactEqCheck-8         	       1	6052688709 ns/op
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-checkWithExactEqCheck-8          	       1	7349380146 ns/op
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-checkWithApproximateEqCheck-8    	       1	10030765581 ns/op
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-majorityWithExactEqCheck-8       	       1	33215526638 ns/op
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-majorityWithApproximateEqCheck-8 	       1	39880077185 ns/op
```
with Boyer–Moore majority vote algorithm:
```
goos: linux
goarch: amd64
pkg: github.com/bookingcom/carbonapi/pkg/backend
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-normalWithExactEqCheck-8         	       1	6213874116 ns/op
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-checkWithExactEqCheck-8          	       1	7012712732 ns/op
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-checkWithApproximateEqCheck-8    	       1	10050420245 ns/op
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-majorityWithExactEqCheck-8       	       1	20160802573 ns/op
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-majorityWithApproximateEqCheck-8 	       1	31473745254 ns/op
```

**Update:**

After some optimization there is a huge performance improvement (The comparison is between before this PR and the latest changes in this PR):
```
name                                                                        old time/op  new time/op  delta
Renders/BenchmarkRenders/ReplicaMatchMode-normalWithExactEqCheck-8          2.36ms ± 0%  2.26ms ± 0%   ~     (p=1.000 n=1+1)
Renders/BenchmarkRenders/ReplicaMatchMode-checkWithExactEqCheck-8           11.1ms ± 0%   3.6ms ± 0%   ~     (p=1.000 n=1+1)
Renders/BenchmarkRenders/ReplicaMatchMode-checkWithApproximateEqCheck-8     11.5ms ± 0%   4.7ms ± 0%   ~     (p=1.000 n=1+1)
Renders/BenchmarkRenders/ReplicaMatchMode-majorityWithExactEqCheck-8        10.2ms ± 0%   3.8ms ± 0%   ~     (p=1.000 n=1+1)
Renders/BenchmarkRenders/ReplicaMatchMode-majorityWithApproximateEqCheck-8  11.5ms ± 0%   3.9ms ± 0%   ~     (p=1.000 n=1+1)
RendersMismatchStorm/ReplicaMatchMode-normalWithExactEqCheck-8               9.15s ± 0%   7.33s ± 0%   ~     (p=1.000 n=1+1)
RendersMismatchStorm/ReplicaMatchMode-checkWithExactEqCheck-8                18.1s ± 0%    8.0s ± 0%   ~     (p=1.000 n=1+1)
RendersMismatchStorm/ReplicaMatchMode-majorityWithExactEqCheck-8             49.6s ± 0%   10.6s ± 0%   ~     (p=1.000 n=1+1)
RendersMismatchStorm/ReplicaMatchMode-checkWithApproximateEqCheck-8          21.3s ± 0%   13.5s ± 0%   ~     (p=1.000 n=1+1)
RendersMismatchStorm/ReplicaMatchMode-majorityWithApproximateEqCheck-8       63.3s ± 0%   26.2s ± 0%   ~     (p=1.000 n=1+1)
RendersStorm/ReplicaMatchMode-normalWithExactEqCheck-8                       6.68s ± 0%   6.92s ± 0%   ~     (p=1.000 n=1+1)
RendersStorm/ReplicaMatchMode-checkWithExactEqCheck-8                        21.2s ± 0%    9.1s ± 0%   ~     (p=1.000 n=1+1)
RendersStorm/ReplicaMatchMode-majorityWithExactEqCheck-8                     21.0s ± 0%    8.8s ± 0%   ~     (p=1.000 n=1+1)
RendersStorm/ReplicaMatchMode-checkWithApproximateEqCheck-8                  20.9s ± 0%    9.6s ± 0%   ~     (p=1.000 n=1+1)
RendersStorm/ReplicaMatchMode-majorityWithApproximateEqCheck-8               21.3s ± 0%   10.3s ± 0%   ~     (p=1.000 n=1+1)

```

### DIFF
--- a/pkg/backend/benchmark_test.go
+++ b/pkg/backend/benchmark_test.go
@@ -101,44 +101,53 @@ func BenchmarkRenders(b *testing.B) {
 	}
 }
 
-func BenchmarkRendersMismatchStorm(b *testing.B) {
-	secPerMonth := int(30 * 24 * time.Hour / time.Second)
-
-	allZeroValues1 := make([]float64, secPerMonth)
-	allZeroValues2 := make([]float64, secPerMonth)
-	allOneValues := make([]float64, secPerMonth)
-	for i := range allOneValues {
-		allOneValues[i] = 1.0
+func createRegularMismatches(backendNumber int, metrics []carbonapi_v2_pb.FetchResponse) []carbonapi_v2_pb.FetchResponse {
+	MismatchFreq := 30000
+	for mIndex := range metrics {
+		mismatchCount := (len(metrics[mIndex].Values) - backendNumber) / MismatchFreq
+		for i := 0; i < mismatchCount; i++ {
+			vIndex := backendNumber + (i * MismatchFreq)
+			metrics[mIndex].Values[vIndex] += 1.0
+		}
 	}
-	metrics := carbonapi_v2_pb.MultiFetchResponse{
-		Metrics: []carbonapi_v2_pb.FetchResponse{
-			carbonapi_v2_pb.FetchResponse{
-				Name:     "foo",
-				Values:   allZeroValues1,
-				IsAbsent: make([]bool, secPerMonth),
-			},
-			carbonapi_v2_pb.FetchResponse{
-				Name:     "foo",
-				Values:   allZeroValues2,
-				IsAbsent: make([]bool, secPerMonth),
-			},
-			carbonapi_v2_pb.FetchResponse{
-				Name:     "foo",
-				Values:   allOneValues,
-				IsAbsent: make([]bool, secPerMonth),
-			},
+	return metrics
+}
+
+func createFullMismatches(backendNumber int, metrics []carbonapi_v2_pb.FetchResponse) []carbonapi_v2_pb.FetchResponse {
+	for mIndex := range metrics {
+		for i := 0; i < len(metrics[mIndex].Values); i++ {
+			metrics[mIndex].Values[i] += float64(backendNumber)
+		}
+	}
+	return metrics
+}
+
+func createSingleBackendMetrics() []carbonapi_v2_pb.FetchResponse {
+	metricsCount := 600
+	metrics := make([]carbonapi_v2_pb.FetchResponse, metricsCount)
+	dpCount := 10800
+	for i := 0; i < 600; i++ {
+		metrics[i] = carbonapi_v2_pb.FetchResponse{
+			Name:     fmt.Sprintf("metric.foo%d", i),
+			Values:   make([]float64, dpCount),
+			IsAbsent: make([]bool, dpCount),
+		}
+	}
+	return metrics
+}
+
+func BenchmarkRendersStorm(b *testing.B) {
+	metricsByBackend := []carbonapi_v2_pb.MultiFetchResponse{
+		{
+			Metrics: createRegularMismatches(1, createSingleBackendMetrics()),
+		},
+		{
+			Metrics: createRegularMismatches(2, createSingleBackendMetrics()),
+		},
+		{
+			Metrics: createRegularMismatches(3, createSingleBackendMetrics()),
 		},
 	}
-	blob, err := metrics.Marshal()
-	if err != nil {
-		b.Fatal(err)
-	}
-
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Content-Type", "application/x-protobuf")
-		w.Write(blob)
-	}))
-	defer server.Close()
 
 	client := &http.Client{}
 	client.Transport = &http.Transport{
@@ -151,7 +160,17 @@ func BenchmarkRendersMismatchStorm(b *testing.B) {
 	}
 
 	backends := make([]Backend, 0)
-	for i := 0; i < 3; i++ {
+	for i := 0; i < len(metricsByBackend); i++ {
+		blob, err := metricsByBackend[i].Marshal()
+		if err != nil {
+			b.Fatal(err)
+		}
+
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Set("Content-Type", "application/x-protobuf")
+			w.Write(blob)
+		}))
+		defer server.Close()
 		bk, err := bnet.New(bnet.Config{
 			Address: server.URL,
 			Client:  client,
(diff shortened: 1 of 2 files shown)


pkg/backend/benchmark_test.go
```diff
@@ -102,14 +102,30 @@ func BenchmarkRenders(b *testing.B) {
 }
 
 func BenchmarkRendersStorm(b *testing.B) {
```
bom-d-van, line 104: The current benchmark only has 3 metrics and the production issue that we are seeing usually have a higher number of metrics.

Maybe we can introduce a benchmark function that has high metric cardinality and relatively high number of metrics.

6000 metrics * 3 replicas * 10800 data points, not sure if it's too much.

bom-d-van:
tips: https://pkg.go.dev/golang.org/x/tools/cmd/benchcmp could be handy here, it prints out delta changes that's more readable.

```
benchcmp old.txt new.txt
benchcmp is deprecated in favor of benchstat: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat
benchmark                                                                                         old ns/op       new ns/op       delta
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-normalWithExactEqCheck-8             6052688709      6213874116      +2.66%
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-checkWithExactEqCheck-8              7349380146      7012712732      -4.58%
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-checkWithApproximateEqCheck-8        10030765581     10050420245     +0.20%
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-majorityWithExactEqCheck-8           33215526638     20160802573     -39.30%
BenchmarkRendersStorm/BenchmarkRendersStorm/ReplicaMatchMode-majorityWithApproximateEqCheck-8     39880077185     31473745254     -21.08%
```

pkg/types/types.go
```diff
@@ -318,36 +318,38 @@ func areFloatsExactlyEqual(a, b float64) bool {
...
-				majorityValue = v
-			}
+	// Boyer–Moore majority vote algorithm
+	i := 0
+	var m float64
+	for _, v := range values {
+		if i == 0 {
+			m = v
+		}
+		if equalityFunc(m, v) {
```
bom-d-van, line 334: this could be changed to `} else if equalityFunc(m, v) {`, less code path to execute.

emadolsky: No it's not a bug. `i` should be 1 after the first iteration. As it is here.

pkg/backend/benchmark_test.go
emadolsky: Yeah that makes more sense.

pkg/types/types.go
bom-d-van: oh I see, that's clever. but i think by merging the if statement, it still might make a difference on the benchmark numbers. also moving out the `if i > valuesCount/2`. sounds a bit micro-optimisation.

emadolsky: Changed.

pkg/types/types.go
```diff
@@ -374,11 +372,13 @@ func mergeMetrics(metrics []Metric, replicaMismatchConfig cfg.RenderReplicaMisma
...
 	metric = metrics[0]
+	valuesForPoint := make([]float64, 0, len(metrics))
+	isMismatchFindConfig := replicaMatchMode != cfg.ReplicaMatchModeNormal
 	for i := range metric.Values {
 		pointExists := !metric.IsAbsent[i]
-		shouldLookForMismatch := replicaMatchMode != cfg.ReplicaMatchModeNormal
+		shouldLookForMismatch := isMismatchFindConfig
 		mismatchObserved := false
-		var valuesForPoint []float64
+		valuesForPoint = valuesForPoint[:0]
```
bom-d-van, line 381: nice catch! this looks like this the true cause. it was allocating small slice objects in a tight loop, which probably trigger higher gc pressures later on?

emadolsky: It had impacts but it was not the main one. The bigger overhead was `equalityFunc` function calls. Making exact equality check inline, had a bigger impact.

bom-d-van: hmm, really? that's a bit surprising.

pkg/backend/benchmark_test.go
emadolsky: made the change.

pkg/types/types.go
```diff
@@ -419,8 +421,8 @@ func mergeMetrics(metrics []Metric, replicaMismatchConfig cfg.RenderReplicaMisma
 
 		mismatches++
 		if replicaMatchMode == cfg.ReplicaMatchModeMajority {
-			majorityValue, majorityCount, err := getPointMajorityValue(valuesForPoint, equalityFunc)
-			if err == nil && majorityCount > len(valuesForPoint)/2 {
+			majorityValue, isMajority, err := getPointMajorityValue(valuesForPoint, equalityFunc)
```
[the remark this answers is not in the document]
emadolsky: It is mostly because of short-circuiting check mode. but if you have any ideas, do not hesitate to suggest.

pkg/types/types.go
bom-d-van: nice!

probably made sense to drop the `equalityFunc` in the future. really no need to support approximate comparison. whisper files has deterministic aggregation/rollup behaviour.

pkg/types/types.go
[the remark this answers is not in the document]
bom-d-van: **this is a lot of back and forth for you, so you don't have to rush adopting the suggestion bellow. if the current optimisation you made can already do the job, it's good enough for me.**

---

my thinking is that we should go all the way back to restore the loop as close to its original behaviour: https://github.com/bookingcom/carbonapi/blob/e237af483315b5262e18af8c57cfdc19831ce0b6/pkg/types/types.go#L249-L272 

To my limited knowledge, with majority healing, the best we can get by using the BM algorithm is still O(3 * N) (3 as in the replica number we have on production and N is the total uniq data points returned in a query), comparing to the current simple null healing.

the code snippet bellow is most likely buggy and incorrect, but just to convey my idea. the allocation of `valuesForPoint` is avoided in nature, and we shall drop approximate check function based on your finding.

```go
metric = metrics[0]
majorityHealing := true
majorityHealed := 0
for i := range metric.Values {
	if !metric.IsAbsent[i] && !majorityHealing {
		continue
	}

	var mcount int
	var mval float64
	var found bool
	for j := 0; j < len(metrics); j++ {
		m := metrics[j]

		if m.StepTime != metric.StepTime || len(m.Values) != len(metric.Values) {
			break
		}

		if m.IsAbsent[i] {
			continue
		}
		found = true

		if mcount == 0 {
			mval = m.Values[i]
			mcount++
		} else if mval == m.Values[i] {
			mcount++
		} else {
			mcount--
		}
	}

	if found {
		metric.IsAbsent[i] = false
		metric.Values[i] = mval
	}
	healed++
	if mcount != len(metrics) {
		majorityHealed++
	}
}
```

bom-d-van (approved):

pkg/types/types.go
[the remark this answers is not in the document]
emadolsky: Thanks a lot for your time! Boyer-Moore algorithm always chooses a value even when there is no majority. And we want to have metrics on unfixed mismatches. so another pass on the values is needed (that's why we need that `valuesForPoint` slice).
I agree that your code snippet is way more readable and I hope we can make it like what you suggested. However, we were trying to give highest performance for each of match configs (normal, check, majority). And this, introduced complexities. Of course we can use the same path for both check & majority but IMO that doesn't make much sense and it's better to drop check mode config. But I think currently it is something that we can postpone.   
About approximate checks, I agree. But still, I cannot be 100% sure that it wouldn't hurt. So, I guess it'd better be handled in another PR in the future.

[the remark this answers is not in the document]
bom-d-van: the code snippet wasn't trying address all the flags we have now, just for conveying the ideas.

yeah, it sounds good to me. you probably missed my bold opening text on the top. it's intended only as a suggestion.

> this is a lot of back and forth for you, so you don't have to rush adopting the suggestion bellow. if the current optimisation you made can already do the job, it's good enough for me.

PR #356, opened 2022-04-05
Upgrade to go-1.18

emadolsky:
Updated the project to work with the latest go release (1.18).


.golangci.yml
```diff
@@ -8,6 +8,13 @@ run:
   # activate after switching to go modules (#131, #190)
   # modules-download-mode: vendor
 
+  # TODO: This is hard-coded here because of https://github.com/golangci/golangci-lint/issues/2649#issuecomment-1070528726
+  # golangci-lint had issues with go1.18 so some of the linters are disabled. As we are not using go1.18 features
```
grzkv, line 12: I have successfully made it work for 1.18 in NT and it actually works in most cases.
Are there some specific blocks why we can't use 1.18? A lot of fixes were done since the post of mentioned comment. Also, we do not need most if not all of the indicated linters.

.golangci.yml
```diff
@@ -145,12 +152,14 @@ linters:
   enable:
```
grzkv, line 152: Normally, it is best to set golangci-lint either by disabling all linters and then only enabling the necessary ones. Or adding necessary ones on top the default. Maybe we can use that approach here.
Currently, the setup seems a bit conflicting.

grzkv (changes requested):
On top of other comments https://github.com/bookingcom/carbonapi#requirements needs updating too.

grzkv:
@emadolsky Have you tested this PR on a live box? Current change of go version can have a significant impact.

.golangci.yml
emadolsky: :+1: simplified it.

emadolsky:
> On top of other comments https://github.com/bookingcom/carbonapi#requirements needs updating too.

updated.

.golangci.yml
grzkv: I'd suggest rather disabling those linters then moving on with version discrepancy. It is a good question if we need those enabled at all. If not, please make a follow-up issue to fix that later.

grzkv (approved):
The issue with version stays, but I'm ok w/ fixing it later on.

bom-d-van (approved):

Issue #52, opened 2019-01-24
Refactor logging to make it more controlled


grzkv:
Currently, there are the following problems with logging:

- loggers are global variables
- there are many loggers and it is hard to know how many
- new loggers can be spawned randomly
- it is hard to control log format
- it is hard to flush and sync loggers

We need to change this to:

- having one (or two) loggers
- pass these around to avoid global state, control the format and mechanics

grzkv:
Logging works differently on `carbonapi` and `zipper`.

1. On `carbonapi` we log init into `stdout`, and log everything else into the file
2. On `zipper` we log errors and warnings to `stdout` and don't log them to the file

azhiltsov:
One more example which is really annoying:
carbonzipper logs: 
```json
{
  "level": "INFO",
  "timestamp": "2019-02-05T14:27:19.471+0100",
  "logger": "access",
  "message": "request served",
  "handler": "render",
  "carbonapi_uuid": "7652d624-c031-4384-8bc2-c96039beb7be",
  "format": "protobuf",
}
```
carbonapi:
```json
{
  "level": "INFO",
  "timestamp": "2019-02-05T14:27:19.527+0100",
  "logger": "access",
  "message": "request served",
  "data": {
    "handler": "render",
    "carbonapi_uuid": "7652d624-c031-4384-8bc2-c96039beb7be",
    .......
  },
}
```
it's hard to impossible to match data.carbonapi_uuid to carbonapi_uuid in ElasticSearch

azhiltsov:
the same query in logs, all of them are legit, but non uniform
carboanpi: "format": "json"
zipper find:   "format": "protobuf",
zipper render:   "format": "protobuf",
go-carbon find:  "format": "protobuf",
go-carbon render:  "format": "carbonapi_v2_pb",

Probably we need some uniformity here, I propose for protobuf to keep the protocol version in the logs.

emadolsky:
In addition to what is already mentioned, log levels are not used pretty well IMO. We are producing ERROR logs on client errors (404s) which makes it hard to distinguish internal issues from typical client errors.

grzkv:
@emadolsky Completely agree about logging levels.

grzkv:
I have re-worded the issue. Now it has a clear set of fixes that we need to resolve it.
@emadolsky Let me know if we need to make any corrections.

grzkv:
@emadolsky Added #359 to the issue description.

PR #357, opened 2022-04-08
Refactor logging

emadolsky:
This PR aims to solve logging issues described on https://github.com/bookingcom/carbonapi/issues/52.

* Flattened carbonapi `data` nest in logs
* Injected loggers from app into handlers and functions
* Got rid of different handler loggers ("render", "find", etc.), and used one handler logger
* Removed CorruptionLogger as it is not used anymore

Tests are changed accordingly and after the first pass of review, it will be tested on production environment.

### DIFF
--- a/app/carbonapi/app.go
+++ b/app/carbonapi/app.go
@@ -398,14 +398,20 @@ func (app *App) deferredAccessLogging(r *http.Request, accessLogDetails *carbona
 
 	accessLogDetails.Runtime = time.Since(t).Seconds()
 	accessLogDetails.RequestMethod = r.Method
+
+	fields, err := accessLogDetails.GetLogFields()
+	if err != nil {
+		accessLogger.Error("could not marshal access log details", zap.Error(err))
+	}
+
 	// TODO (grzkv) This logic is not obvious for the user
 	if logAsError {
-		accessLogger.Error("request failed", zap.Any("data", *accessLogDetails))
+		accessLogger.Error("request failed", fields...)
 		apiMetrics.Errors.Add(1)
 	} else {
 		// TODO (grzkv) The code can differ from the real one. Clean up
 		// accessLogDetails.HttpCode = http.StatusOK
-		accessLogger.Info("request served", zap.Any("data", *accessLogDetails))
+		accessLogger.Info("request served", fields...)
 		apiMetrics.Responses.Add(1)
 	}
 
--- a/app/carbonapi/app.go
+++ b/app/carbonapi/app.go
@@ -181,7 +181,7 @@ func setUpConfig(app *App, logger *zap.Logger) {
 		}
 	}
 
-	functions.New(app.config.FunctionsConfigs)
+	functions.New(app.config.FunctionsConfigs, logger)
 
 	// TODO (grzkv): Move expvars to init since they are global to the package
 	expvar.Publish("config", expvar.Func(func() interface{} { return app.config }))
--- a/app/carbonapi/app.go
+++ b/app/carbonapi/app.go
@@ -101,12 +101,14 @@ func New(config cfg.API, logger *zap.Logger, buildVersion string) (*App, error) 
 // Start starts the app: inits handlers, logger, starts HTTP server
 func (app *App) Start() func() {
 	logger := zapwriter.Logger("carbonapi")
+	accessLogger := zapwriter.Logger("access")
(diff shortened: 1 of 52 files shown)


grzkv:
@emadolsky I think the #52 can be considered fully solved when we are left with only one logger. But I agree that this is a large refactoring that is better done in several steps.
Also, I think it makes sense to make #52 more specific. I'll do some edits there.

pkg/handler_log/log.go
```diff
@@ -0,0 +1,14 @@
+package handler_log
```
grzkv, line 1: It is uncommon to use underscores in the names of files, packages, folders.

app/carbonapi/app.go
```diff
@@ -101,12 +101,14 @@ func New(config cfg.API, logger *zap.Logger, buildVersion string) (*App, error)
 // Start starts the app: inits handlers, logger, starts HTTP server
 func (app *App) Start() func() {
 	logger := zapwriter.Logger("carbonapi")
+	accessLogger := zapwriter.Logger("access")
```
grzkv, line 104: Does this actually reduce the number of loggers and move us clser to having a single logger?

grzkv: Or is this maybe rerquired as a preliminary step?

grzkv (changes requested):

expr/functions/absolute/function_test.go
```diff
@@ -17,7 +17,7 @@ func init() {
 	metadata.SetEvaluator(evaluator)
 	helper.SetEvaluator(evaluator)
 	for _, m := range md {
-		metadata.RegisterFunction(m.Name, m.F)
+		metadata.RegisterFunction(m.Name, m.F, nil)
```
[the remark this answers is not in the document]
emadolsky: legit.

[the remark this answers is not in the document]
emadolsky: Used nop.

app/carbonapi/app.go
emadolsky: With the latest changes we have 3 loggers:
* main: stdout logger in the main
* access: access logging
* app (carbonapi/zipper): setup, handlers

We can merge app and access but I think it's reasonable to have them separated.

grzkv: I am pretty sure we need just *one* logger. I do not see why we'd need more.

grzkv: That is the strategy of majority of projects. It adds uniformity and clarity. There should be a good reason to have more than one.

emadolsky: At the current state, loggers are controlled. Both loggers are initialized on app.Start and passed through the handlers. No magic, no global logger, etc.
The one in the main can be eliminated by using "log" instead of "zap". But I don't think that is a big deal.
For the other two, we have the separation in the code. Access logs are separated from logic logs and they can be configured and used differently.
If we merge this two, we do not get much benefit IMO.

grzkv: What benefit do we get from having two loggers? Why not have 10?
For me, I don't understand why we should have a separate access log. Neither is it clear when should I use it.

If we do not need two loggers, added complexity is not necessary. I think that should be obvious. This seems like a wrong design decision to me that is the leftovers of the problem that we are trying to solve.

grzkv (changes requested):
We need to end up with one logger for the issue to be fully resolved. Otherwise, this seems like a partial resolution.

emadolsky:
> @emadolsky Is there a reason we continue using the `zapwriter`? If we do logging properly, it should not be necessary. Unless we use some functionality that cannot be replicated with `zap` itself.

@grzkv we are using zapwriter config in corbonapi/zipper config and it needs some more changes. I propose we migrate to zap in another PR. WDYT?

grzkv:
> > @emadolsky Is there a reason we continue using the `zapwriter`? If we do logging properly, it should not be necessary. Unless we use some functionality that cannot be replicated with `zap` itself.
> 
> @grzkv we are using zapwriter config in corbonapi/zipper config and it needs some more changes. I propose we remove zapwriter in another PR. WDYT?

Agree. Made #359

grzkv:
As a note, the logs remain global as long as we continue using `zapwriter`. And global access to logging is probably it's biggest problem. As long as `zapwriter` stays, #52 is mostly unresolved.

grzkv (approved):

PR #361, opened 2022-05-09
Break globs in carbonapi render

emadolsky:

Breaking the render requests having more than `MaxBatchSize` hits into queries with single metric, causes a huge amount of requests to backends. One solution to this issue is to try and break the big glob into smaller globs but not exact metrics.

~~In this PR, I proposed an assumption on the uniformity of the queries. Therefore, by looking at the number of unique values for each star (or containing star) namespace, we estimate the number of requests avoided and the number of matches per each query. But for being safe, we also check the results and try different error rates.~~
After checking with more real queries, for some cases, the former approach didn't work. So, a new approach is used. The new greedy approach is that we start from the star namespace with more unique members and try to see how many results we will have by placing the star there. If we don't exceed the limit, we continue. Else, we will just move on to the next ns.
It is not an optimal solution, but it doesn't hurt performance and reduces the number of requests.

Some small tests are provided. Also it's tested against real big queries.


grzkv:
@emadolsky Maybe it's better to put this code into some proper place instead of the `types` package that needs to be refactored. This only makes the problem of the `types` package worse.

PR #364, opened 2022-05-19
Refactor log levels

emadolsky:
Client errors were producing error level logs which didn't make sense. This PR revises the log levels in carbonapi and carbonzipper.

### DIFF
--- a/app/carbonapi/app.go
+++ b/app/carbonapi/app.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"expvar"
 	"fmt"
+	"go.uber.org/zap/zapcore"
 	"net"
 	"net/http"
 	"os"
@@ -383,7 +384,7 @@ func setUpConfig(app *App, logger *zap.Logger) {
 
 }
 
-func (app *App) deferredAccessLogging(accessLogger *zap.Logger, r *http.Request, accessLogDetails *carbonapipb.AccessLogDetails, t time.Time, logAsError bool) {
+func (app *App) deferredAccessLogging(accessLogger *zap.Logger, r *http.Request, accessLogDetails *carbonapipb.AccessLogDetails, t time.Time, level zapcore.Level) {
 	accessLogDetails.Runtime = time.Since(t).Seconds()
 	accessLogDetails.RequestMethod = r.Method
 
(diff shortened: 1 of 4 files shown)


app/carbonapi/app.go
```diff
@@ -391,15 +391,20 @@ func (app *App) deferredAccessLogging(accessLogger *zap.Logger, r *http.Request,
 	if err != nil {
 		accessLogger.Error("could not marshal access log details", zap.Error(err))
 	}
-
+	var logMsg string
```
grzkv, line 394: This is a change in logic. Now we can call the function to log an error but it will be logged as "request served".
Also, not all requests that all not 2xx are failures. 
Is this what we want?

grzkv (commented):
@emadolsky There are a lot of errors that are now qualified as info. Could you please explain the reasoning behind this?

app/carbonapi/app.go
emadolsky: > Could you please explain the reasoning behind this?
Yes, of course. As I already mentioned in the PR description, we are getting lots of error logs caused by client errors. Error level should simply be for the cases we experience an issue inside the system.
As an example, if we cannot parse results from go-carbon, it is an error. If we get timed out while gathering the results, it is an error. But if the user provides a wrong value for a field, it's not.

> Now we can call the function to log an error but it will be logged as "request served"
Yes you can. But we never do in the code even though it seems right when we have written the response with 2xx to log "request served". It just doesn't make sense to call the function like that.
Anyways, the whole way of logging with this `logAsError` is not desirable in my opinion. But I didn't want to have a bigger change in the logging logic in carbonapi. This can be done in another PR or some following commits (I prefer the former). 

> Also, not all requests that all not 2xx are failures.
From what I observed, all non-2xx cases are failures in the code (we only have 2xx, 4xx, and 5xx, don't we?).

app/carbonapi/http_handlers.go
```diff
@@ -587,7 +583,7 @@ func (app *App) renderHandlerProcessForm(r *http.Request, accessLogDetails *carb
 	if tstr := r.FormValue("cacheTimeout"); tstr != "" {
 		t, err := strconv.ParseInt(tstr, 10, 64)
 		if err != nil {
-			logger.Error("failed to parse cacheTimeout",
+			logger.Info("failed to parse cacheTimeout",
```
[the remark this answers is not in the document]
grzkv: I agree that should not be an error if we take the approach you mentioned. However, I think it's best if we log this as warning https://github.com/uber-go/zap/blob/master/level.go#L36 
What do you think?

app/carbonapi/app.go
grzkv: > Also, not all requests that all not 2xx are failures.
This is fundamentally true. Logging otherwise is treating HTTP codes wrongfully.

IMO, we can remove `logMsg` altogether. Opinions?

grzkv:
@emadolsky I'd suggest to log client errors as warnings as a general approach. Frequently they can be an indication that something is wrong, so the level of warning is more fit for that.

I totally agree that logging client errors as errors can interfere with observability.

app/carbonapi/app.go
emadolsky: I'm not sure what you're proposing. What will we log then?

emadolsky:
@grzkv I understand but I believe there are other indicators when something is wrong. Like, if we have a drop on success rate of the requests or an increase on client errors. If we do log them as warning, we will be logging a lot of warnings which really do not mean warning at all, and we don't even look at them as something wrong cause they're going to be there the whole time.
So, I still believe that client errors in requests should be treated as INFO. It just doesn't make sense at all to log warnings when there is a wrong query string in the URL.
For some special cases like 499, I agree that it should be logged differently and that is the reason they have become errors in this PR. I'm gonna change those to WARN.

grzkv (approved):

PR #370, opened 2022-07-04
Zipper CarbonV2 gRPC render stream

emadolsky:
This PR adds the functionality of gRPC streaming Renders in carbonzipper. The new gRPC protocol is proposed at https://github.com/go-graphite/protocol/pull/12, and storage part is implemented on https://github.com/go-graphite/go-carbon/pull/476.

#### Outline:

A new backend configuration structure is proposed as `ProtocolBackend` which defines different protocol addresses of a backend. The change is backward compatible.

The current implementation only uses gRPC for Renders as we have not implemented other APIs here or in go-carbon. For the other APIs such as Find, old methods are used.

The changes are tested manually in container environment.

Some works should be done before merging:
- [ ] More tests on real-life workload


cfg/common_test.go
```diff
@@ -187,8 +187,8 @@ backendsByCluster:
       - "http://10.190.202.32:8080"
       - "http://10.190.197.92:8080"
       protocolBackends:
-      - address: "http://10.190.198.92:7070"
-        grpc: true
+      - http: "http://10.190.198.93:8080"
+        grpc: "10.190.198.93:7070"
```
bom-d-van, line 191: I guess we would need to note that it's not backward compatible in the pr due to the config changes.

pkg/backend/grpc/grpc.go
```diff
@@ -0,0 +1,14 @@
...
+)
+
+// Backend represents a host that accepts requests for metrics over gRPC.
+type Backend struct {
+	net.Backend
+}
+
+//func (b Backend) Render(ctx context.Context, request types.RenderRequest) ([]types.Metric, error) {
+//
+//}
```
bom-d-van: dead code?

pkg/backend/net/grpc.go
```diff
@@ -0,0 +1,97 @@
...
+			return nil, err
+		}
+		fetchedMetrics = append(fetchedMetrics, types.Metric{
+			Name:      fetchResponse.Name,
+			StartTime: fetchResponse.StartTime,
+			StopTime:  fetchResponse.StopTime,
+			StepTime:  fetchResponse.StepTime,
+			Values:    fetchResponse.Values,
+			IsAbsent:  fetchResponse.IsAbsent,
+		})
```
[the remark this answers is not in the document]
emadolsky: Before streaming, what we did was to send the requests (which could have 1 to 10k metrics) to the backends, and merge them afterwards.
It is still the case as we do not aim to fix that in this specific PR.
I agree that it can be more optimized to merge the metrics on the go, but that needs some interface changes in the code and it can get a bit complex. Right now, It's not necessary to implement all this. It is something that we can consider after having streaming. WDYT?

cfg/common_test.go
emadolsky: Actually it is. the older `backends` still works. This new list is called `protocolBackends`.

pkg/backend/grpc/grpc.go
emadolsky: It is deleted in the following commits.

pkg/backend/net/grpc.go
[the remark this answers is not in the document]
bom-d-van: oh, so we still do split with streaming in this pr? I see, if it's gonna come as a separate change, then I guess yes. 

I was dreaming resolving the big query split with streaming. :P

cfg/common_test.go
bom-d-van: oh, I see. you know what, we actually go with 'grpc://10.190.198.93:7070' and skip a new config field. but just a wild idea.

pkg/backend/net/grpc.go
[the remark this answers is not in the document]
emadolsky: Yeah I believe we should do that in the end but I'm trying to keep the PRs smaller. Like for example gRPC is only implemented for Render ATM.

Generally, I think the following works after this should be:
* Implementing other rpcs as well and fully migrate to gRPC.
* Parallelize merging metrics with the Render stream for better performance and memory consumption
* Merging zipper and carbonapi for better performance and memory consumption 

As a result of these, we can remove the breaking logic.

cfg/common_test.go
emadolsky: Well I thought about this but the issue is that we should have a fall back strategy for find, info, etc. as they are not yet implemented in gRPC. So with the current changes, Renders hit the gRPC endpoint and others use the older implementation. Although, I am open to change the plan to implement all of these rpcs prior to going on production but that will require more changes, hence, more time and risk.

cfg/common_test.go
```diff
@@ -374,6 +393,13 @@ backendsByDC:
           - name: "cluster2"
             backends:
             - "http://10.290.202.32:8080"
+    - name: "dc3"
+      clusters:
+          - name: "cluster1"
+            protocolBackends:
+            - http: "http://10.291.202.31:8080"
+            - http: "http://10.291.197.91:8080"
+              grpc: "10.291.197.91:7070"
```
bom-d-van, line 402: it's minor, but you can simplify things by:

```
protocolBackends:
- "http://10.291.202.31:8080"
- "http://10.291.197.91:8080"
  "grpc://10.291.197.91:7070"
```

net/url pkg actually supports parsing it: https://pkg.go.dev/net/url@go1.18.4#Parse

bom-d-van (approved):
left one more small comment. but not a blocker and the rest seems good to me, so approved.

(I will leave the real check to production. :P)

cfg/common_test.go
emadolsky: I have difficulty understanding the for `protocolBackend`. Do you mean it can be simplified to list of strings? In that case it would be like:
```protocolBackends:
- ["http://10.291.202.31:8080"]
- ["http://10.291.197.91:8080",
  "grpc://10.291.197.91:7070"]```

Which is not desired IMO. However, after implementing gRPC for all of the calls, it can be simplified to one string in which url scheme determines the backend type.

PR #371, opened 2022-07-08
Refactor sendGlobsAsIs

### LINKED ISSUE #33: Refactor sendGlobAsIs ?
We have 3 parameters:
sendGlobsAsIs: true|false
alwaysSendGlobsAsIs: true|false
maxBatchSize: int

Their logic is convoluted 
**sendGlobsAsIs** in _true_ is working together with **maxBatchSize**
 carbonapi is sending a find request and depending on amount of metrics returned from stores acts differently:
 * if amount of metrics is lower than maxBatchSize it sends request as is with globs
 * if amount of metrics is higher it uses list of metrics returned from find and query them one by one

**sendGlobsAsIs** in _false_ is always sending find and ignoring **maxBatchSize**

**alwaysSendGlobsAsIs** I suppose is neglecting **sendGlobsAsIs** settings and always send a render request without 'find'

Proposing deprecate **sendGlobsAsIs** and **alwaysSendGlobsAsIs** in favor of
**resolveGlobs:** true|false in conjunction with **maxBatchSize**

resolveGlobs: false -> send render query as it is
resolveGlobs: true -> send find query and if:
  count of metrics < maxBatchSize - send as it is
  count of metrics > maxBatchSize - group them in batches of maxBatchSize and send them in batches

azhiltsov:
We have 3 parameters:
sendGlobsAsIs: true|false
alwaysSendGlobsAsIs: true|false
maxBatchSize: int

Their logic is convoluted
sendGlobsAsIs in true is working together with maxBatchSize
carbonapi is sending a find request and depending on amount of metrics returned from stores acts differently:

if amount of metrics is lower than maxBatchSize it sends request as is with globs
if amount of metrics is higher it uses list of metrics returned from find and query them one by one
sendGlobsAsIs in false is always sending find and ignoring maxBatchSize

alwaysSendGlobsAsIs is neglecting sendGlobsAsIs settings and always send a render request without 'find'

This commit is deprecating sendGlobsAsIs maxBatchSize and alwaysSendGlobsAsIs in favor of
resolveGlobs: int

resolveGlobs: = 0  -> send render query as it is
resolveGlobs: = 1  -> send find request and send render for every metric separately
resolveGlobs: > 1  -> send find query and if:
count of metrics < resolveGlobs - send render as it is
count of metrics > resolveGlobs - send render query for each metric returned from the find query

Close #33

### DIFF
--- a/app/carbonapi/http_handlers.go
+++ b/app/carbonapi/http_handlers.go
@@ -5,14 +5,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/bookingcom/carbonapi/pkg/handlerlog"
-	"go.uber.org/zap/zapcore"
 	"net/http"
 	"runtime/debug"
 	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
+
+	"github.com/bookingcom/carbonapi/pkg/handlerlog"
+	"go.uber.org/zap/zapcore"
 
 	"github.com/bookingcom/carbonapi/carbonapipb"
 	"github.com/bookingcom/carbonapi/date"
@@ -164,7 +165,7 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request, logger *za
 		}
 		//2xx response code is treated as success
 		if toLog.HttpCode/100 == 2 {
-			if toLog.TotalMetricCount < int64(app.config.MaxBatchSize) {
+			if toLog.TotalMetricCount < int64(app.config.ResolveGlobs) {
 				app.prometheusMetrics.RenderDurationExpSimple.Observe(time.Since(t0).Seconds())
 				app.prometheusMetrics.RenderDurationLinSimple.Observe(time.Since(t0).Seconds())
 			} else {
@@ -700,11 +701,11 @@ func (app *App) renderWriteBody(results []*types.MetricData, form renderForm, r 
 }
 
 func (app *App) sendGlobs(glob dataTypes.Matches) bool {
-	if app.config.AlwaysSendGlobsAsIs {
+	if app.config.ResolveGlobs == 0 {
 		return true
+	} else {
+		return len(glob.Matches) < app.config.ResolveGlobs
 	}
-
-	return app.config.SendGlobsAsIs && len(glob.Matches) < app.config.MaxBatchSize
 }
 
 func (app *App) resolveGlobsFromCache(metric string) (dataTypes.Matches, error) {
@@ -759,7 +760,7 @@ func (app *App) resolveGlobs(ctx context.Context, metric string, useCache bool, 
 
 func (app *App) getRenderRequests(ctx context.Context, m parser.MetricRequest, useCache bool,
 	toLog *carbonapipb.AccessLogDetails) ([]string, error) {
-	if app.config.AlwaysSendGlobsAsIs {
+	if app.config.ResolveGlobs == 0 {
 		return []string{m.Metric}, nil
 	}
 	if !strings.ContainsAny(m.Metric, "*{") {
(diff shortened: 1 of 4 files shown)


app/carbonapi/http_handlers.go
```diff
@@ -759,7 +762,7 @@ func (app *App) resolveGlobs(ctx context.Context, metric string, useCache bool,
 
 func (app *App) getRenderRequests(ctx context.Context, m parser.MetricRequest, useCache bool,
 	toLog *carbonapipb.AccessLogDetails) ([]string, error) {
-	if app.config.AlwaysSendGlobsAsIs {
+	if app.config.ResolveGlobs <= 0 {
```
bom-d-van, line 765: this should be `== 0` now?

app/carbonapi/http_handlers.go
```diff
@@ -700,11 +701,13 @@ func (app *App) renderWriteBody(results []*types.MetricData, form renderForm, r
 }
 
 func (app *App) sendGlobs(glob dataTypes.Matches) bool {
-	if app.config.AlwaysSendGlobsAsIs {
+	if app.config.ResolveGlobs == 0 {
```
timtofan, line 704: PR description says
> resolveGlobs: <= 0 -> send render query as it is

But this code enables find upon resolveGlobs=0?

bom-d-van (approved):

app/carbonapi/http_handlers.go
```diff
@@ -759,7 +762,7 @@ func (app *App) resolveGlobs(ctx context.Context, metric string, useCache bool,
 
 func (app *App) getRenderRequests(ctx context.Context, m parser.MetricRequest, useCache bool,
 	toLog *carbonapipb.AccessLogDetails) ([]string, error) {
-	if app.config.AlwaysSendGlobsAsIs {
+	if app.config.ResolveGlobs == 0 {
```
timtofan, line 765: should this be `<= 0`?

bom-d-van: no, this one should be `== 0`, I just commented it above. :P

config/carbonapi.yaml
```diff
@@ -29,29 +29,37 @@ cache:
...
+# alwaysSendGlobsAsIs: true|false
+# maxBatchSize: int
+# Migration path 
+# alwaysSendGlobsAsIs: true -> resolveGlobs: 0
+# sendGlobsAsIs: false -> resolveGlobs: -1
+# sendGlobsAsIs: true && maxBatchSize: int -> resolveGlobs: int
+
+resolveGlobs: 100
+# = 0 - faster (no 'find in advance', direct render)
+# < 0 - slower (always 'find in advance' and every metric rendered individually)
```
timtofan, line 47: This seems to contradict description in PR?
> resolveGlobs: <= 0 -> send render query as it is

timtofan, line 47: Would it make sense to change it from `<0` to simply `1`? Meaning resolveGlobs=1  enforces /metrics/find and separate /render for each metric. This way the behaviour could be clear w/o reading description. While for `<0` behaviour is not self-evident w/o reading docs.

timtofan:
Seems like a new way is way simpler 👍  Left a few comments.

app/carbonapi/http_handlers.go
timtofan: I guess intended behaviour is a bit unclear atm cause PR description doesn't match explanation in comments. Let's wait for those to sync up and then prolly this thread will be clearer.

app/carbonapi/http_handlers.go
timtofan: after description got fixed, this is no longer relevant

config/carbonapi.yaml
emadolsky: I agree with @timtofan here. With your proposal, negative values and 1 will result in the same way. So why bother define a new condition after all?

azhiltsov: Understood. This is a quite elegant solution!

azhiltsov (commented):
I think it ready for the final review

PR #372, opened 2022-07-20
Use noproxy dial option on grpc client

emadolsky:
For http backend client, we do not respect the http proxy environment, so we should do the same on gRPC.


azhiltsov (approved):
LGTM

PR #378, opened 2022-08-24
Added per-backend histogram for finds in zipper. Upgraded prometheus.

grzkv:
Have better visibility for `find` requests in `zipper`.

1. We need to keep the number of exposed metrics manageable in case there are many backends. This is why we use an exponential histogram. Linear one would not cover enough range.
2. Prometheus summaries aren't a good option. They won't work across a fleet of instances; we cannot aggregate them.

It was tested on live boxes.


app/carbonzipper/metrics.go
```diff
@@ -189,6 +190,17 @@ func NewPrometheusMetrics(config cfg.Zipper) *PrometheusMetrics {
 					config.Monitoring.FindDurationLin.BucketsNum),
 			},
 		),
+		FindDurationByBackend: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name: "find_duration_seconds_by_backend",
```
timtofan, line 195: I think unit part (`_seconds`) usually go last, this would also make metric name closer to struct field name. Though not sure what's the pattern in carbonzipper, if any.

timtofan: ok seems like surrounding code puts it also in the middle..

timtofan (approved):

auguzun (approved):

azhiltsov (approved):

PR #385, opened 2022-09-07
Switched to a single labeled metric for the time in q in zipper. Start of zipper redesign.

grzkv:
Note: This PR includes unfinished refactoring. Some rough edges are to be expected.

* Fixes time in queue for zipper stats.
* Refactoring/redesign of the zipper.
  * Removed parts of the global state
  * Partially removed expvars
  * Restructured modules
  * Misc cleanups

### DIFF
--- a/pkg/backend/net/net.go
+++ b/pkg/backend/net/net.go
@@ -50,9 +50,10 @@ type Backend struct {
 	client         *http.Client
 	timeout        time.Duration
 	limiter        *prioritylimiter.Limiter
-	logger         *zap.Logger
 	cache          *expirecache.Cache
 	cacheExpirySec int32
+
+	logger         *zap.Logger
 }
 
 // Config configures an HTTP backend.
@@ -297,7 +298,7 @@ func (b Backend) Render(ctx context.Context, request types.RenderRequest) ([]typ
 	u = carbonapiV2RenderEncoder(u, from, until, targets)
 	request.Trace.AddMarshal(t0)
 
-	contentType, resp, err := b.call(ctx, request.Trace, u)
+	contentType, resp, err := b.call(ctx, request.Trace, u, )
 	if err != nil {
 		if code, ok := err.(ErrHTTPCode); ok && code == http.StatusNotFound {
 			return nil, types.ErrMetricsNotFound
--- a/pkg/backend/net/net.go
+++ b/pkg/backend/net/net.go
@@ -53,7 +53,8 @@ type Backend struct {
 	cache          *expirecache.Cache
 	cacheExpirySec int32
 
-	logger         *zap.Logger
+	qHist  *prometheus.HistogramVec // note: we do not use zipper's metrics to separate from capi
+	logger *zap.Logger
 }
 
 // Config configures an HTTP backend.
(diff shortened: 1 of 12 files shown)


pkg/backend/net/net.go
```diff
@@ -238,17 +243,25 @@ func (b Backend) do(trace types.Trace, req *http.Request) (string, []byte, error
...
-func (b Backend) call(ctx context.Context, trace types.Trace, u *url.URL) (string, []byte, error) {
+func (b Backend) call(ctx context.Context, trace types.Trace, u *url.URL, request string) (string, []byte, error) {
 	ctx, cancel := b.setTimeout(ctx)
 	defer cancel()
 
-	t0 := time.Now()
+	var t *prometheus.Timer
+	if b.qHist != nil { // TODO: remove condition when capi is merged with zipper
+		t = prometheus.NewTimer(b.qHist.WithLabelValues(request))
+		t.ObserveDuration()
```
emadolsky, line 253: I think we shouldn't observe duration here.

pkg/backend/net/net.go
```diff
@@ -238,17 +243,25 @@ func (b Backend) do(trace types.Trace, req *http.Request) (string, []byte, error
...
+	}
+
 	err := b.enter(ctx)
-	trace.AddLimiter(t0)
 	if err != nil {
 		return "", nil, err
 	}
 
+	if t != nil {
```
emadolsky, line 261: If I'm not mistaken, having this after the error check, we'll miss the times in queues which lead to error (e.g. timeout). I think we want to observe those times as well.

pkg/backend/net/net.go
```diff
@@ -370,7 +383,7 @@ func (b Backend) Info(ctx context.Context, request types.InfoRequest) ([]types.I
 	u = carbonapiV2InfoEncoder(u, metric)
 	request.Trace.AddMarshal(t0)
 
-	_, resp, err := b.call(ctx, request.Trace, u)
+	_, resp, err := b.call(ctx, request.Trace, u, "info")
```
emadolsky, line 386: Why don't we keep `trace.AddLimiter` and use the time stored there after `b.call` to observe the time on histogram?
This way, we omit this `request string` argument which looks a bit weird IMHO.

pkg/backend/net/net.go
grzkv: Yep, we shouldn't.

pkg/backend/net/net.go
grzkv: I don't think we should. This is intentional but debatable.

pkg/backend/net/net.go
grzkv: `trace` will go away, no need to keep it. Also, it averages stats which is wrong.

pkg/backend/net/net.go
grzkv: Pushed. Thanks for finding it!

emadolsky (approved):
We discussed the PR with Roman, and as there are a lot of things that will still change in the process, we decided to continue, and do not bother ATM.

PR #389, opened 2022-09-13
Polished e2e system test. Added to makefile.

grzkv:
* test runs on Mac now
* test is not CI-system-specific
* corrected shell style and improved safety
* added the makefile target

### DIFF
--- a/tests/system_test.sh
+++ b/tests/system_test.sh
@@ -6,44 +6,22 @@
 #
 # Needs to be run from the root directory of the project.
 
-# Use `begin_fold` and `end_fold` to fold large pieces of output in the job log in Travis CI
-begin_fold() {
-    if [ "$TRAVIS" == true ]; then
-        echo "travis_fold:start:$1"
-    fi
-    echo -e "\e[33;1m$2\e[0m"
-}
-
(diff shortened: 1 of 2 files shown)


tests/system_test.sh
```diff
@@ -71,16 +47,13 @@ for i in `seq 1 $CARBAPI_RETRIES`; do
     fi
 done
 
-if [ $i -eq $CARBAPI_RETRIES ]; then
+if [ "$i" -eq $CARBAPI_RETRIES ]; then
```
timtofan, line 50: `-eq` is for numeric comparison, hence quoting "$i" looks dodgy, since all it could achieve compared to having no quotes is that it keeps whitespaces, which is not something numeric comparison would like

grzkv: This is driven by the shellcheck warning.

grzkv: More specifically, as you say, it prevents splitting. This should not happen, of course, but this is simply defensive programming. If it does happen, it's better with quotes than without. I think this is why the linter insists it is a warning.

timtofan (approved):

PR #390, opened 2022-09-13
Cleaned-up obsolete parts. Moved testdata to the appropriate package.

grzkv:
see title


auguzun (approved):

deniszh (approved):
Agreed, clickhouse probably will work with upstream carbonapi only and not really relevant for us. Not sure about graphite template, but can be restored if needed

PR #394, opened 2022-09-21
Add tld cache extra prefixes to zipper

emadolsky:
This new config adds the functionality to probe deeper paths on zipper in order to involve less backends on queries when metrics are broken into different cluster from the non-leading namespaces.

In addition to normal TLD probes, probes with the configured prefixes are also sent. the results already gathered from the previous probes sent at the same iteration is used to optimize the backends receiving the probes.

New tests are introduced and the functionality is manually tested on local.

### DIFF
--- a/pkg/app/zipper/http_handlers.go
+++ b/pkg/app/zipper/http_handlers.go
@@ -570,7 +570,7 @@ func (app *App) lbCheckHandler(w http.ResponseWriter, req *http.Request, ms *Pro
 func (app *App) filterBackendByTopLevelDomain(targets []string) []backend.Backend {
 	targetTlds := make([]string, 0, len(targets))
 	for _, target := range targets {
-		targetTlds = append(targetTlds, getTopLevelDomain(target))
+		targetTlds = append(targetTlds, getTopLevelDomain(target, app.Config.TLDCacheExtraPrefixes))
 	}
 
 	bs := app.filterByTopLevelDomain(app.Backends, targetTlds)
@@ -580,8 +580,19 @@ func (app *App) filterBackendByTopLevelDomain(targets []string) []backend.Backen
 	return app.Backends
 }
 
-func getTopLevelDomain(target string) string {
-	return strings.SplitN(target, ".", 2)[0]
+func getTopLevelDomain(target string, extraPrefixes []string) string {
+	tld := strings.SplitN(target, ".", 2)[0]
+	for _, prefix := range extraPrefixes {
+		if strings.HasPrefix(target, prefix) {
+			nsCount := strings.Count(prefix, ".") + 1
+			splitTarget := strings.SplitN(target, ".", nsCount+2)  // prefix + ns | rest
+			foundTLD := strings.Join(splitTarget[:nsCount+1], ".") // prefix + ns
+			if len(foundTLD) > len(tld) {
+				tld = foundTLD
+			}
+		}
+	}
+	return tld
 }
 
 func (app *App) filterByTopLevelDomain(backends []backend.Backend, targetTLDs []string) []backend.Backend {
--- a/pkg/app/zipper/http_handlers.go
+++ b/pkg/app/zipper/http_handlers.go
@@ -18,7 +18,6 @@ import (
 	"net/http"
 	"sort"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/bookingcom/carbonapi/pkg/backend"
@@ -570,7 +569,7 @@ func (app *App) lbCheckHandler(w http.ResponseWriter, req *http.Request, ms *Pro
 func (app *App) filterBackendByTopLevelDomain(targets []string) []backend.Backend {
 	targetTlds := make([]string, 0, len(targets))
 	for _, target := range targets {
-		targetTlds = append(targetTlds, getTopLevelDomain(target, app.Config.TLDCacheExtraPrefixes))
+		targetTlds = append(targetTlds, getTargetTopLevelDomain(target, app.TLDPrefixes))
 	}
 
 	bs := app.filterByTopLevelDomain(app.Backends, targetTlds)
@@ -578,21 +577,6 @@ func (app *App) filterBackendByTopLevelDomain(targets []string) []backend.Backen
 		return bs
 	}
 	return app.Backends
-}
-
-func getTopLevelDomain(target string, extraPrefixes []string) string {
-	tld := strings.SplitN(target, ".", 2)[0]
-	for _, prefix := range extraPrefixes {
-		if strings.HasPrefix(target, prefix) {
-			nsCount := strings.Count(prefix, ".") + 1
-			splitTarget := strings.SplitN(target, ".", nsCount+2)  // prefix + ns | rest
-			foundTLD := strings.Join(splitTarget[:nsCount+1], ".") // prefix + ns
-			if len(foundTLD) > len(tld) {
-				tld = foundTLD
-			}
-		}
-	}
-	return tld
 }
 
 func (app *App) filterByTopLevelDomain(backends []backend.Backend, targetTLDs []string) []backend.Backend {
--- a/pkg/app/zipper/tldcache.go
+++ b/pkg/app/zipper/tldcache.go
@@ -6,22 +6,24 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+
 	"github.com/bookingcom/carbonapi/pkg/backend"
 	"github.com/bookingcom/carbonapi/pkg/types"
-	"github.com/pkg/errors"
 )
+
+type tldPrefix struct {
+	prefix        string
+	segments      []string
+	segmentsCount int
+}
 
 func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
 	probeTicker := time.NewTicker(time.Duration(app.Config.InternalRoutingCache) * time.Second) // TODO: The ticker resources are never freed
 	for {
 		topLevelDomainCache := make(map[string][]*backend.Backend)
-		// We should always have empty prefix to get default TLDs
-		allPrefixes := []string{""}
-		allPrefixes = append(allPrefixes, app.Config.TLDCacheExtraPrefixes...)
-		// Sorting to avoid involving unnecessary backends
-		sortedPrefixes := sortedByNsCount(allPrefixes)
-		for _, prefix := range sortedPrefixes {
-			prefix = trimPrefix(prefix)
+		for _, prefix := range app.TLDPrefixes {
 			bs := getBackendsForPrefix(prefix, app.Backends, topLevelDomainCache)
 			for i := range bs {
 				topLevelDomains, err := getTopLevelDomains(*bs[i], prefix)
@@ -45,39 +47,19 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
 	}
 }
 
-type tldPrefix struct {
-	prefix  string
-	nsCount int
-}
-
-func sortedByNsCount(prefixes []string) []string {
-	countedPrefixes := make([]tldPrefix, len(prefixes))
-	for i, prefix := range prefixes {
-		nsCount := 0
-		if prefix != "" {
-			nsCount = strings.Count(prefix, ".") + 1
-		}
-		countedPrefixes[i] = tldPrefix{
-			prefix:  prefix,
-			nsCount: nsCount,
-		}
+func (p *tldPrefix) query() string {
+	query := "*"
+	if p.prefix != "" {
+		query = p.prefix + ".*"
 	}
-	sort.Slice(countedPrefixes, func(i, j int) bool {
-		return countedPrefixes[i].nsCount < countedPrefixes[j].nsCount
-	})
-	sortedPrefixes := make([]string, len(prefixes))
-	for i := range countedPrefixes {
-		sortedPrefixes[i] = countedPrefixes[i].prefix
-	}
-	return sortedPrefixes
+	return query
 }
 
 // getBackendsForPrefix returns the backends that need to be queried in order to populate TLD cache for the prefix.
-// it reuses already fetched tlds to find out about the info. if no info is there, it returns all the backends.
-func getBackendsForPrefix(prefix string, backends []backend.Backend, tldCache map[string][]*backend.Backend) []*backend.Backend {
-	segments := strings.Split(prefix, ".")
-	for i := len(segments); i > 0; i-- {
-		p := strings.Join(segments[:i], ".")
+// It reuses already fetched tlds to find out about the info. If no info is there, it returns all the backends.
+func getBackendsForPrefix(prefix tldPrefix, backends []backend.Backend, tldCache map[string][]*backend.Backend) []*backend.Backend {
+	for i := prefix.segmentsCount; i > 0; i-- {
+		p := strings.Join(prefix.segments[:i], ".")
 		if filteredBackends, ok := tldCache[p]; ok {
 			return filteredBackends
 		}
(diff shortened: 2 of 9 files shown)


pkg/app/zipper/tldcache.go
```diff
@@ -13,16 +15,25 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
-			}
-			for _, topLevelDomain := range topLevelDomains {
-				topLevelDomainCache[topLevelDomain] = append(topLevelDomainCache[topLevelDomain], &app.Backends[i])
+		// We should always have empty prefix to get default TLDs
+		allPrefixes := []string{""}
+		allPrefixes = append(allPrefixes, app.Config.TLDCacheExtraPrefixes...)
+		// Sorting to avoid involving unnecessary backends
+		sortedPrefixes := sortedByNsCount(allPrefixes)
+		for _, prefix := range sortedPrefixes {
+			prefix = trimPrefix(prefix)
```
timtofan, line 24: imho it's better to never override loop variable for better readability

It will work, but the one who reads the code needs to know that range copies the value, and "original" value won't be changed. I think depending on knowing such details doesn't help readability.

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,68 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+	nsCount int
+}
+
+func sortedByNsCount(prefixes []string) []string {
+	countedPrefixes := make([]tldPrefix, len(prefixes))
+	for i, prefix := range prefixes {
+		if prefix == "" {
+			countedPrefixes[i] = tldPrefix{
+				prefix:  prefix,
+				nsCount: 0,
```
timtofan, line 59: nit: imho it would be simpler to calculate nsCount first, and have single code path for assignment

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,68 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+			break
+		}
+		lastDotIndex := strings.LastIndex(prefix, ".")
+		if lastDotIndex == -1 {
+			for i := range backends {
+				bs = append(bs, &backends[i])
+			}
+			break
+		}
+		prefix = trimPrefix(prefix[:lastDotIndex])
```
timtofan, line 91: imho we shouldn't be modifying input arg, but have a separate variable for mutation so that it's easier to see lifecycle of the variable

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,68 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+	sortedPrefixes := make([]string, len(prefixes))
+	for i := range countedPrefixes {
+		sortedPrefixes[i] = countedPrefixes[i].prefix
+	}
+	return sortedPrefixes
+}
+
+func getBackendsForPrefix(prefix string, backends []backend.Backend, tldCache map[string][]*backend.Backend) []*backend.Backend {
+	var bs []*backend.Backend
+	for {
```
timtofan, line 79: imho eternal loop is a dangerous construct, and generally should be avoided

maybe we could structure the code w/o it 🤔

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,68 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+		if prefixBackends, ok := tldCache[prefix]; ok {
+			bs = prefixBackends
+			break
```
timtofan, line 82: maybe those 2 lines can be replaced by return 🤔

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,68 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
 // Returns the backend's top-level domains.
-func getTopLevelDomains(backend backend.Backend) ([]string, error) {
+func getTopLevelDomains(backend backend.Backend, prefix string) ([]string, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	request := types.NewFindRequest("*")
+	query := "*"
+	if prefix != "" {
+		query = prefix + ".*"
```
timtofan, line 107: nit: we could also do it like this
```
	request := types.NewFindRequest(strings.join([]string{prefix,"*"}, ".")
```

dunno which approach is clearer though 🤔

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,68 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+		return countedPrefixes[i].nsCount < countedPrefixes[j].nsCount
+	})
+	sortedPrefixes := make([]string, len(prefixes))
+	for i := range countedPrefixes {
+		sortedPrefixes[i] = countedPrefixes[i].prefix
+	}
+	return sortedPrefixes
+}
+
+func getBackendsForPrefix(prefix string, backends []backend.Backend, tldCache map[string][]*backend.Backend) []*backend.Backend {
```
timtofan, line 77: Imho this func is doing something unexpected. It's called getBackendsForPrefix, yet when there is no match it's not returning empty array. Maybe getBackendsForPrefixOrAll might be more intuitive 🤔

pkg/app/zipper/tldcache.go
```diff
@@ -13,16 +15,25 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
-				topLevelDomainCache[topLevelDomain] = append(topLevelDomainCache[topLevelDomain], &app.Backends[i])
+		// We should always have empty prefix to get default TLDs
+		allPrefixes := []string{""}
+		allPrefixes = append(allPrefixes, app.Config.TLDCacheExtraPrefixes...)
+		// Sorting to avoid involving unnecessary backends
+		sortedPrefixes := sortedByNsCount(allPrefixes)
+		for _, prefix := range sortedPrefixes {
+			prefix = trimPrefix(prefix)
+			bs := getBackendsForPrefix(prefix, app.Backends, topLevelDomainCache)
+			for i := 0; i < len(bs); i++ {
```
timtofan, line 26: nit: maybe we could use range() here? it's a tad simpler and is kinda default usually

pkg/app/zipper/tldcache.go
```diff
@@ -13,16 +15,25 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+				topLevelDomains, err := getTopLevelDomains(*bs[i], prefix)
```
timtofan, line 27: it's not introduced by you, but since you are touching that code:
I think it's a little weird to use longer topLevelDomain all over the place given we already use TLD abbreviation in package name and other places. IMHO we should be consistent in using either short or long version. And in this particular case, imho short version would be preferred since it's already established acronym in config and package name.

pkg/app/zipper/tldcache.go
emadolsky: actually this ends up creating a string starting with "." if prefix is "". So I thought the condition was needed.

timtofan: ah 🤦

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,64 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+	})
+	sortedPrefixes := make([]string, len(prefixes))
+	for i := range countedPrefixes {
+		sortedPrefixes[i] = countedPrefixes[i].prefix
+	}
+	return sortedPrefixes
+}
+
+// getBackendsForPrefix returns the backends that need to be queried in order to populate TLD cache for the prefix.
+// it reuses already fetched tlds to find out about the info. if no info is there, it returns all the backends.
```
timtofan, line 76: nit: if it's formatted as proper sentences w/ dots, then might be a good idea to also start the sentences from capital letter

pkg/cfg/common.go
```diff
@@ -177,9 +177,10 @@ type Common struct {
 	KeepAliveInterval         time.Duration `yaml:"keepAliveInterval"`
 	MaxIdleConnsPerHost       int           `yaml:"maxIdleConnsPerHost"`
 
-	ExpireDelaySec             int32 `yaml:"expireDelaySec"`
-	InternalRoutingCache       int32 `yaml:"internalRoutingCache"`
-	GraphiteWeb09Compatibility bool  `yaml:"graphite09compat"`
+	ExpireDelaySec             int32    `yaml:"expireDelaySec"`
+	InternalRoutingCache       int32    `yaml:"internalRoutingCache"`
+	TLDCacheExtraPrefixes      []string `yaml:"tldCacheExtraPrefixes"`
```
grzkv, line 182: If I understand correctly, we can add prefixes that have more than one level of hierarchy. For example, we can have `a.b.c`, `a.b.d`, and `hello` here.

We may consider using a tree of strings here instead of a list. So, we will go from:

```
tldCacheExtraPrefixes:
  - "a.b.c"
  - "a.b.d"
  - "hello"
```
to for example
```
tldCacheExtraPrefixes:
  - "a":
      "b":
        - "c"
        - "d"
  - "hello"
```
(of course, there are also other ways to specify hierarchy, like with a nested list).

This will be more type-safe. It would make parsing a bit harder but then the other parsing and sorting logic will become much simpler.

At the same time, we do not expect many levels of hierarchy. So, it will still be convenient when we want to have e.g. a single prefix `a.b.c`.

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,64 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+		}
+	}
+	allBackends := make([]*backend.Backend, len(backends))
+	for i := range backends {
+		allBackends[i] = &backends[i]
+	}
+	return allBackends
+}
+
+func trimPrefix(prefix string) string {
```
grzkv, line 92: A minor thing. This seems like a proxy function. Please consider removing it.

pkg/app/zipper/tldcache.go
```diff
@@ -13,16 +15,25 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
-				// lg.Error("failed to probe TLD cache for a backend", zap.Error(err), zap.String("backend", app.Backends[i].GetServerAddress()))
-				ms.TLDCacheProbeErrors.Inc()
-			}
-			for _, topLevelDomain := range topLevelDomains {
-				topLevelDomainCache[topLevelDomain] = append(topLevelDomainCache[topLevelDomain], &app.Backends[i])
+		// We should always have empty prefix to get default TLDs
+		allPrefixes := []string{""}
+		allPrefixes = append(allPrefixes, app.Config.TLDCacheExtraPrefixes...)
+		// Sorting to avoid involving unnecessary backends
+		sortedPrefixes := sortedByNsCount(allPrefixes)
```
grzkv, line 22: We pre-process the prefixes configuration - sorting, trimming, etc. - every time we probe the TLDs. If I'm not mistaken, we don't need to do this repeatedly. We only need to do it once.

I'd suggest to move these operations our of this loop and into a separate function for config processing. We can also have the validation there later on.

pkg/app/zipper/tldcache.go
```diff
@@ -34,12 +45,64 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+	for i := range countedPrefixes {
+		sortedPrefixes[i] = countedPrefixes[i].prefix
+	}
+	return sortedPrefixes
+}
+
+// getBackendsForPrefix returns the backends that need to be queried in order to populate TLD cache for the prefix.
+// it reuses already fetched tlds to find out about the info. if no info is there, it returns all the backends.
+func getBackendsForPrefix(prefix string, backends []backend.Backend, tldCache map[string][]*backend.Backend) []*backend.Backend {
+	segments := strings.Split(prefix, ".")
```
grzkv, line 78: I'm not sure we need to do this every time. As mentioned in another comment, this could be moved to a separate function to process prefixes config.

pkg/app/zipper/tldcache.go
```diff
@@ -13,16 +15,25 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
-			topLevelDomains, err := getTopLevelDomains(app.Backends[i])
-			if err != nil {
-				// this could add a lot of noise to logs
-				// lg.Error("failed to probe TLD cache for a backend", zap.Error(err), zap.String("backend", app.Backends[i].GetServerAddress()))
-				ms.TLDCacheProbeErrors.Inc()
-			}
-			for _, topLevelDomain := range topLevelDomains {
-				topLevelDomainCache[topLevelDomain] = append(topLevelDomainCache[topLevelDomain], &app.Backends[i])
+		// We should always have empty prefix to get default TLDs
+		allPrefixes := []string{""}
```
grzkv, line 19: I think we may consider storing the list of prefixes as a parsed, trimmed, and sorted tree of strings (it will more trie actually). The operations will become much simpler. The code will be more clear. We won't need to do processing repeatedly. The performance will be better. There will be less memory allocation. More importantly, the validation will be backed-in.

@emadolsky What do you think?

grzkv (dismissed):
I have left a number of comments. We can merge this as-is and address them later.
So, approving this for now to unblock even though there are still questions pending.
@emadolsky I'm happy to discuss the questions now if you think that's more appropriate.

pkg/cfg/common.go
emadolsky: Yes it can be implemented like this. however, I don't see a huge benefit from implementing it this way.
Also, it'll create and ambiguity that when we configure:
```tldCacheExtraPrefixes:
  - "a":
      "b":
        - "c"
        - "d"
  - "hello"```
will it induce probing of `a` or `a.b` or is it only `a.b.c` and `a.b.d`.
Although it might not be a normal use-case, probing `a.b.c` should not necessarily cause probing of `a.b`. Of course we can add some attributes to each prefix to identify if it should be probed, but I believe that is not neat.

File: pkg/app/zipper/tldcache.go [hunk 16]
```diff
@@ -13,16 +15,25 @@ func (app *App) probeTopLevelDomains(ms *PrometheusMetrics) {
...
+		allPrefixes = append(allPrefixes, app.Config.TLDCacheExtraPrefixes...)
```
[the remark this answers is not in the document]
emadolsky: Yes I agree. invalid inputs probably do not cause big harms but still it's nice to have. I was even thinking about keeping and using a validated sorted list of prefixes so we wouldn't need to sort it every time. Also it'll make the query lookups faster. 
We can either do it now or later. However, if I do it right now, it's likely that it should be moved and changed during your refactor. So, your call.

pkg/app/zipper/tldcache.go
emadolsky: Yeah exactly. I described the same thing on a previous comment.

pkg/app/zipper/tldcache.go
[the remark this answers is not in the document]
emadolsky: I already made the changes. please check again - @grzkv.

pkg/app/zipper/tldcache.go
emadolsky: It's the name of the function. Capitalizing it, make it wrong.

timtofan: I meant "It reuses .." and "If no info is there .." sentences.

emadolsky: Oh sorry. Fixed.

pkg/cfg/common.go
grzkv: @emadolsky Yes, there are pros and cons to both. I'm totally ok if we leave this like it is.

grzkv (approved):

pkg/app/zipper/http_handlers.go
```diff
@@ -570,7 +569,7 @@ func (app *App) lbCheckHandler(w http.ResponseWriter, req *http.Request, ms *Pro
 func (app *App) filterBackendByTopLevelDomain(targets []string) []backend.Backend {
 	targetTlds := make([]string, 0, len(targets))
 	for _, target := range targets {
-		targetTlds = append(targetTlds, getTopLevelDomain(target))
+		targetTlds = append(targetTlds, getTargetTopLevelDomain(target, app.TLDPrefixes))
```
timtofan, line 572: nit: we are using both Tld and TopLevelDomain in same code, imho we could standardize on one of the two for consistency, maybe Tld since it's used also in package name and config_

pkg/app/zipper/tldcache.go
```diff
@@ -50,3 +87,56 @@ func getTopLevelDomains(backend backend.Backend) ([]string, error) {
...
+func InitTLDPrefixes(logger *zap.Logger, cfgPrefixes []string) []tldPrefix {
+	tldPrefixes := []tldPrefix{
+		// We should always have empty prefix to get default TLDs
+		{prefix: "", segments: nil, segmentsCount: 0},
+	}
+
+	for _, p := range cfgPrefixes {
+		segments := strings.Split(p, ".")
+		var invalid bool
+		for _, s := range segments {
```
timtofan, line 102: might be worth to move this into a func, and add tests for it

timtofan: also IIUC validation could be also done via strings.Contains(p,"..")

Issue #136, opened 2019-05-03
Remove expvar setup.


grzkv:
Following the recommendation [in official docs](https://golang.org/pkg/expvar/#Publish).

`expvar`s are global and should be registered globally. We should also consider trimming the content and moving non-system info to Prometheus endpoints.

grzkv:
Changed this to remove the expvars. They are superseded by the Prometheus metrics.

PR #395, opened 2022-09-22
Removed duplicated and obsolete expvars metrics. Removed unused pathcache.

grzkv:
* Removed unused pathcache abstraction.
* Removed most of the expvars. They are either duplicated in Prometheus or unused (or both). Partially addressed #136
* Further removed global state.

Local and real-life test runs.


emadolsky:
Thanks for the huge cleaning!
If I'm not mistaken, for some of the graphite metrics removed, we don't have Prometheus alternative.
e.g. `apiMetrics.FindCacheHits` can be a valuable metric that we'll lose.
So I propose adding alternatives either in this PR or in a following PR. WDYT? @grzkv

grzkv:
@emadolsky There are a couple of metrics that are not duplicated. However, we need to keep in mind, that the Graphite and expvars metrics are mostly unused. So, we don't lose anything in terms of observability at the moment.

We're currently implementing additional metrics and will add the missing ones along the way. It is likely that we will not duplicate them exactly one-to-one. E.g. it is better to have `totalCacheRequests` + `cacheMisses` instead of `cacheHits` + `cacheMisses`. Also, the separation of metrics is off sometimes: e.g. we will likely merge some of them as a single metric with different labels. 

In other words, it is best not to duplicate one-to-one but instead to implement the missing metrics so that they fall into the general picture properly. This is the approach I will try to follow.

So, in my opinion, we should do the implementation of metrics separately.

PR #398, opened 2022-09-27
Fixing out of index error during metric sorting

deniszh:
It's related to this PR https://github.com/go-graphite/carbonapi/issues/128. Before this change it's possible that query can have more dots than requested metric. Obvious example:
Metrics: 
```
a.b1.c.d
a.b2.c.d
a.b3.c.d
a.b4.c.d
```
Query: `a.{*.b1,*}.c.*`
Query looks strange, but it's valid and it was taken from real example. 
So, in the case above sorting will cause out of index error in `getPart()` function (and caused panic, intercepted only in http module, which is not good).

Instead of splitting query by dots using simple `strings.Split()` I'm splitting it by using tricky regex, which will ignore dots inside square or curly bracket pairs. In that case sort will work properly and will not cause errors in queries above. 
I also included range protection in getPart() but proper splitting should fix underlying issue and not just mask it, as range protection alone doing. Performance wise that should not do much hurt - regex compiles once, split will be done once for every query, only if it contains globs. 

I added 2 more test cases for sort - one for error which I fixed and another to test that sort by glob also happening.

### DIFF
--- a/expr/sort.go
+++ b/expr/sort.go
@@ -2,6 +2,7 @@ package expr
 
 import (
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -31,8 +32,15 @@ func (b byPartBase) Swap(i, j int) {
 }
 
 func getPart(metric *types.MetricData, part int) string {
+	// metric name should not contain braces
+	// we can use normal split here
 	parts := strings.Split(metric.Name, ".")
-	return parts[part]
+	// returning proper part if in range
+	if part < len(parts) {
+		return parts[part]
+	}
+	// otherwise return last part
+	return parts[len(parts)-1]
 }
 
 // Given two indices, i and j, and a comparator function that returns whether
@@ -114,13 +122,19 @@ func sortByBraces(metrics []*types.MetricData, part int, pattern string) {
 	}
 }
 
+// compile regex for sorting
+// it splits string by dots, but ignoring enclosed {} and []
+// it fixes bug when number of dots in pattern and metric can differ
+// causing out of index error in getParts() above
+var reSplitByDotsIgnoreBraces = regexp.MustCompile(`\s*(\{[^}]*\}|\[[^]]*\]|[^.]+)`)
+
 // SortMetrics sort metric data alphabetically.
 func SortMetrics(metrics []*types.MetricData, mfetch parser.MetricRequest) {
 	// Don't do any work if there are no globs in the metric name
 	if !strings.ContainsAny(mfetch.Metric, "*?[{") {
 		return
 	}
-	parts := strings.Split(mfetch.Metric, ".")
+	parts := reSplitByDotsIgnoreBraces.FindAllString(mfetch.Metric, -1)
 	// Proceed backwards by segments, sorting once for each segment that has a glob that calls for sorting.
 	// By using a stable sort, the rightmost segments will be preserved as "sub-sorts" of any more leftward segments.
 	for i := len(parts) - 1; i >= 0; i-- {
--- a/expr/sort.go
+++ b/expr/sort.go
@@ -121,9 +121,9 @@ func sortByBraces(metrics []*types.MetricData, part int, pattern string) {
 	}
 }
 
-// splitByDotsIgnoringBraces split string by dots, ignoring dots in curly and normal braces
+// SplitByDotsIgnoringBraces split string by dots, ignoring dots in curly and normal braces
 // please note it doesn't support enclosed brackets because we don't need it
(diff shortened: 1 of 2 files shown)


emadolsky:
I was thinking about the case in which we had two other metrics:
```
a.f1.b1.c.d
a.f1.b1.c.a
```
With current implementation, these two results are not sorted out by the last part.

expr/sort.go
```diff
@@ -114,13 +122,19 @@ func sortByBraces(metrics []*types.MetricData, part int, pattern string) {
 	}
 }
 
+// compile regex for sorting
+// it splits string by dots, but ignoring enclosed {} and []
+// it fixes bug when number of dots in pattern and metric can differ
+// causing out of index error in getParts() above
+var reSplitByDotsIgnoreBraces = regexp.MustCompile(`\s*(\{[^}]*\}|\[[^]]*\]|[^.]+)`)
```
grzkv, line 129: It is not a good idea to use a global variable unless there are good reasons for it. Sorry, but I don't see a reason to use a global variable here. Please consider making this variable local to where it's used.

I guess, it is global to prevent repeated re-compilation. I don't think this is a sufficient reason to have it as a global.

On top of that, since this is not a simple constant declaration, this will include unexpected and implicit code execution during initialization. This can create many problems.

grzkv, line 129: I can't tell how much this would impact performance in this particular case, but regexs are very slow in Go, especially complex ones like this.

From what I understand, this can be done without regex by using a simple processing function and it would be much faster. This would also eliminate the need for pre-compilation.

On top of that, in my subjective opinion, it would be much easier to understand if it was a function instead of a regex.

deniszh: @grzkv : Having regex in global variables is common pattern, common enough to have it [excluded](https://github.com/leighmcculloch/gochecknoglobals/issues/6) in [gochecknoglobals](https://github.com/leighmcculloch/gochecknoglobals) even.
But in this case it's probably fine, it will be called only once per target, so, I'll make it local.

deniszh: @grzkv: you're probably right, but I'm doubt that I would be able to write function which will be faster than regex.

deniszh: Myeah. Will try to write that function, at least for readability.

grzkv: > doubt that I would be able to write function which will be faster than regex

If I get this correctly, this looks like an O(N) check. It will almost certainly be faster and clearer than regex. See also https://github.com/golang/go/issues/26623

The fact that we've agreed on making the regex local makes this less acute. Global regex was the main reason I was bothering you with this) Still, it would be nice to have.

deniszh: I implemented split function, so, probably not relevant anymore.

deniszh: I wrote split function, please review.

deniszh:
@grzkv @emadolsky : I wrote split function and test, please re-review.

expr/sort.go
```diff
@@ -31,8 +31,15 @@ func (b byPartBase) Swap(i, j int) {
...
 func getPart(metric *types.MetricData, part int) string {
+	// metric name should not contain braces
+	// we can use normal split here
 	parts := strings.Split(metric.Name, ".")
-	return parts[part]
+	// returning proper part if in range
+	if part < len(parts) {
+		return parts[part]
+	}
+	// TODO: that should never happen, maybe we should log it
```
grzkv, line 41: If it's clear that this execution path is wrong and shouldn't happen, it makes sense to return an error. That would be safer: we will make the unexpected behavior evident, and the error will clearly indicate what is wrong.

grzkv (approved):
Approved, even though there are some minor comments on my side.
@deniszh It's your call whether to address them now or later.

expr/sort.go
deniszh: Good idea, but I'm not sure how to do so - as you can see `getPart` is used in `sort.Stable(AlphabeticallyByPart(metrics, i))` and `sort.Stable` do not generate or propagate errors. I will mention that in TODO, though.

expr/sort.go
```diff
@@ -114,13 +121,35 @@ func sortByBraces(metrics []*types.MetricData, part int, pattern string) {
 	}
 }
 
+// SplitByDotsIgnoringBraces split string by dots, ignoring dots in curly and normal braces
+// please note it doesn't support enclosed brackets because we don't need it
+func SplitByDotsIgnoringBraces(str string) []string {
```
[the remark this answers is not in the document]
deniszh: Ah, mea culpa. It's better to unexport it. Done, please re-review.

grzkv (approved):

PR #402, opened 2022-09-30
Improve capi cache logic and logs

emadolsky:
Currently, there's no visibility over cache errors. This PR takes into account errors happened during set/get of cache.

It is ensured that the errors are escalated and caught on references. Then, errors are added to carbonapi request logs.

Tests are improved.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -5,7 +5,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"strings"
-	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/cache/cache.go
```diff
@@ -109,10 +108,9 @@ func (m *MemcachedCache) Get(k string) ([]byte, error) {
 	return item.Value, nil
 }
 
-func (m *MemcachedCache) Set(k string, v []byte, expire int32) {
-	key := sha256.Sum256([]byte(k))
-	hk := hex.EncodeToString(key[:])
-	go m.client.Set(&memcache.Item{Key: m.prefix + hk, Value: v, Expiration: expire})
+func (m *MemcachedCache) Set(k string, v []byte, expire int32) error {
+	hk := getCacheHashKey(k)
+	return m.client.Set(&memcache.Item{Key: m.prefix + hk, Value: v, Expiration: expire})
```
deniszh, line 113: Set will be sync after that change, right? I know that's easier for error but will set timeout work in that case?

auguzun (approved):

pkg/cache/cache.go
emadolsky: Set function of other caches are sync. It made no sense to have an inconsistency over the types implementing the same interface. So, I made this consistent with others (sync).

deniszh: Ok, makes sense then.

deniszh (approved):

grzkv (approved):

PR #405, opened 2022-10-04
Observability additions for carbonapi.

### LINKED ISSUE #241: Add Prometheus metrics for cache usage in carbonapi
The current metrics coverage for caching behavior in `carbonapi` is poor.

We need to add sensible caching metrics to `carbonapi that would reflect the behavior of caches.

grzkv:
Several observability additions:

* a set of metrics for carbonapi memcached. Fixes #241 
* metrics to count requests propagated to remotes from carbonapi
* added metrics to the zipper limiter in carbonapi

Also, added a limiter to find requests from carbonapi to zipper.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/bradfitz/gomemcache/memcache"
-	"github.com/google/go-cmp/cmp"
 )
 
 type TestMemcache struct {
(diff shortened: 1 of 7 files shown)


pkg/cache/cache_test.go
```diff
@@ -42,70 +41,70 @@ func (m *TestMemcache) Set(i *memcache.Item) error {
...
-	bData := []byte("bval")
-	m.Set("a", aData, 0)
-	m.Set("b", bData, 0)
-
-	aRes, err := m.Get("a")
-	if err == nil || err == ErrNotFound {
-		t.Fatalf("Expected timeout, got val %v, err %v", aRes, err)
-	}
-}
+// func TestReplicatedMemcacheWithPartialTimeout(t *testing.T) {
```
timtofan, line 44: Out of curiosity, why the tests got commented out?

grzkv: Because I have removed redundant polymorphism. We cannot have this test without it.

grzkv: @timtofan This is the sacrifice we will have to make to make the code more clear and get access to other memcache ops moving forward. 
In general, in my experience, it is usually not a great idea to unit-test things that require remote connections. We can do it, but the overhead is not worth it.

grzkv: @timtofan We have also discussed this with @emadolsky and agreed not to do unit testing when it has high overhead. In my opinion, this is one such case.

grzkv: I think, it makes more sense to remove the tests, not comment them out. So, I did that.

timtofan: thanks for the info

timtofan (approved):

deniszh (approved):

PR #409, opened 2022-10-12
Added logs in zipper when metrics schema mismatch

auguzun:
Logs will allow to track metrics with schema mismatch

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -267,7 +267,7 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
 	}
 	if len(diffResolutionMetrics) > 0 {
 		logger.Info("metric with different schema observed",
-			zap.Any("schema_mismatch_metrics", diffResolutionMetrics),
+			zap.Strings("schema_mismatch_metrics", diffResolutionMetrics),
 			zap.Int("metrics total", len(diffResolutionMetrics)),
 		)
 	}
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -276,10 +276,9 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
 }
 
 func isStepTimeMatching(ms []Metric) bool {
-	if len(ms) != 0 {
-		defaultStep := ms[0].StepTime
-		for _, mr := range ms {
-			if defaultStep != mr.StepTime {
+	if len(ms) > 1 {
+		for i := 1; i < len(ms); i++ {
+			if ms[i].StepTime != ms[0].StepTime {
 				return false
 			}
 		}
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -233,8 +233,12 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
 		MismatchedPoints int    `json:"mismatched_points"`
 	}
 	var mismatchedMetricReports []metricReport
+	var diffResolutionMetrics []string
 	for _, ms := range metricByNames {
 		m, stats := mergeMetrics(ms, replicaMismatchConfig)
+		if !isStepTimeMatching(ms) {
+			diffResolutionMetrics = append(diffResolutionMetrics, m.Name)
+		}
 		unfixedMismatches := stats.MismatchCount - stats.FixedMismatchCount
 		if unfixedMismatches > 0 &&
 			len(mismatchedMetricReports) < replicaMismatchConfig.RenderReplicaMismatchReportLimit {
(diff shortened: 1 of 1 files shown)


pkg/types/types.go
```diff
@@ -261,10 +265,28 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
 			zap.Int("replica_points_total", metricsStat.DataPointCount),
 		)
 	}
+	if len(diffResolutionMetrics) > 0 {
+		logger.Info("metric with different schema observed",
+			zap.Any("schema_mismatch_metrics", diffResolutionMetrics),
```
emadolsky, line 270: I think we can use zap.Strings here which is better.

emadolsky (changes requested):

emadolsky (approved):

pkg/types/types.go
```diff
@@ -261,10 +265,28 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
 			zap.Int("replica_points_total", metricsStat.DataPointCount),
 		)
 	}
+	if len(diffResolutionMetrics) > 0 {
```
grzkv, line 268: Do we know how frequently this happens in real life? If this happens a lot, this may significantly impact the volume of logs. Have we tested for this?

Have we considered adding a metric instead of a log record? Do we need the details in the log record, or would a metric suffice?

pkg/types/types.go
```diff
@@ -261,10 +265,28 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
...
+		logger.Info("metric with different schema observed",
+			zap.Strings("schema_mismatch_metrics", diffResolutionMetrics),
+			zap.Int("metrics total", len(diffResolutionMetrics)),
+		)
+	}
 
 	return merged, metricsStat
 }
 
+func isStepTimeMatching(ms []Metric) bool {
```
grzkv, line 278: Have we checked the performance impact of this? Is the performance impact justified to have this information?

pkg/types/types.go
```diff
@@ -233,8 +233,12 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
 		MismatchedPoints int    `json:"mismatched_points"`
 	}
 	var mismatchedMetricReports []metricReport
+	var diffResolutionMetrics []string
 	for _, ms := range metricByNames {
 		m, stats := mergeMetrics(ms, replicaMismatchConfig)
+		if !isStepTimeMatching(ms) {
+			diffResolutionMetrics = append(diffResolutionMetrics, m.Name)
```
grzkv, line 240: This can result in a high number of allocations. I'm a bit concerned about the performance impact of this. BTW, using a metric would resolve this problem.

pkg/types/types.go
```diff
@@ -261,10 +265,28 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
...
+			zap.Strings("schema_mismatch_metrics", diffResolutionMetrics),
+			zap.Int("metrics total", len(diffResolutionMetrics)),
+		)
+	}
 
 	return merged, metricsStat
 }
 
+func isStepTimeMatching(ms []Metric) bool {
+	if len(ms) != 0 {
```
grzkv, line 279: I guess we should have `> 1` here. No point to check when there's a single metric.

pkg/types/types.go
```diff
@@ -261,10 +265,28 @@ func MergeMetrics(metrics [][]Metric, replicaMismatchConfig cfg.RenderReplicaMis
...
+		defaultStep := ms[0].StepTime
+		for _, mr := range ms {
```
grzkv, line 281: Since we will have the guarantee of `len(ms) > 1` with the above requested change, we can simplify this to something like this
```
for i := 1; i < len(ms); i++ {
    if mr.StepTime != mr[0].StepTime {
        return false
        ...
```

pkg/types/types.go
deniszh: 👍 for live test, impact can be significant, agreed

pkg/types/types.go
deniszh: @grzkv : I'm not an expert here, but is variable (`defaultStep`) comparison in loop not better then comparison with structure element of slice (`mr[0].StepTime`)? Or compiler optimise that anyway?

pkg/types/types.go
auguzun: I tested on live box and didn't see perf degradation

pkg/types/types.go
grzkv: @deniszh Interesting question! There's no difference, I think. Both will be a single address in assembly. The compiler _may_ need to re-calculate the address in the second case, but I'm not sure, it may also optimize it away. Even if it does not, the difference is zero in practice.

pkg/types/types.go
auguzun: changed, thanks

pkg/types/types.go
auguzun: changed, thanks

pkg/types/types.go
auguzun: changed, thanks

grzkv (approved):

Issue #407, opened 2022-10-07
Redesign the request processing pipeline.


grzkv:
There are a number of problems with how we process requests:

1. The limiter has some problems by itself:
    * It has no limit on the number of concurrent requests. This sometimes results in high number of concurrent goroutines that compete for a single channel lock. This also creates a potential positive feedback loop as contention reduces performance, that in turn increases contention, and so on.
    * Context cancellation is the only way to remove a request from the limiter. Cancellations are unpredictable and hard to reason about.
    * The previous two points at times of high load introduce high contention that reduces performance when it's needed most.
2. There is no fairness in how the requests are processed. See #35
3. There is no global limit on the number of requests sent to backends in zipper. Since the limits for individual backends need to be generous (because the connections are extremely unevenly distributed), the total number of connections can skyrocket introducing critical overhead. This can result — and we have seen this — in positive feedback loop involving remote caches. Such occurrence can make the system unstable.
4. It is often unclear how the current processing works because its components don't have a clear separation of responsibilities. This makes diagnostics especially hard.

We need to address the mentioned issues with a redesign.

Issue #35, opened 2019-01-16
Unlucky requests can get needlessly dropped


grzkv:
Our requests limiter works like a semaphore now. The requests are not processed in a FIFO queue but are picked randomly from an unordered pool. The requests have a timeout. This means, that some requests can get *unlucky* and will not be picked up for longer than needed, and will be timed out.

## Example
Say, we have `10 requests` to be processed, each takes `1 second`, but every second we get a new request in. The number of requests to be processed remains constant i.e. `10`. Requests are processed one-at-a-time.

Intuitively, the waiting time for a request should be `9 seconds`.

### What we have now
Requests are picked randomly. In this case, the probability that a request is picked is `0.9` at each processing cycle. The chance that a request will be in the queue for >30 seconds is `0.9^29 ~ 5%`. This is much longer than needed, and chances are that the request will be timed out and dropped.

### What would be nice to have
Requests go into a FIFO queue. This way, each request waits for `9 seconds`, then it is processed. The waiting time is more predictable and fair.

grzkv:
Here's an illustration of what happens https://play.golang.org/p/V4riLAtcUAq

ksurent:
If you're looking to optimise throughput in scenarios with many concurrent requests, there're [some](https://queue.acm.org/detail.cfm?id=2839461) [systems](https://blogs.dropbox.com/tech/2018/03/meet-bandaid-the-dropbox-service-proxy/) out there that adopted LIFO queues. Could be interesting to see if this strategy can be useful here.

grzkv:
@ksurent Thanks for the suggestions. We will be able to compare after we do the refactoring.

I think, in this case, we need something simple. Just a classic limited-length requests queue instead of a semaphore. Or something like that.

ksurent:
Sure, I was just suggesting a different queueing discipline (LIFO vs FIFO).

gunnihinn:
You actually don't want a FIFO queue.

Under normal, happy conditions it doesn't matter what type of queue you have because almost everything gets served.

When you're under heavy load and need to start shedding some to survive, using a FIFO queue can cause you to time out almost all the requests sitting in your queue if you pick a few heavy requests from it in a row. If you instead pick requests randomly, or using a FILO queue, you increase the expected number of requests you can serve even under that load.

IIRC, the first Google SRE book has some discussion of this in one of the chapters on load shedding or cascading failures. Andre Vereira at Booking also knows some stuff about this if you want a second face-to-face opinion.

[...]

PR #412, opened 2022-10-19
First part of the new processing system.

grzkv:
This PR contains the top level processor of the new request processing system together with the necessary observability. This is an intermediate stage of refactoring and the design will be finished in the follow-up PRs.

Please check the comments in the code for technical details.

This is the first stage of solving the #407 and #35

### Migration and next steps
The old limiter is kept in place for now. This way we can be sure the transition is smooth. It will be removed during the next stages.

The following steps will also introduce the new backend limiter in `zipper`.

### Test results
This was extensively tested locally and on live boxes. The tests show that the latency performance is at least as good as before.

At the same time, **the peak number of goroutines is reduced approximately x20 times**.

### DIFF
--- a/pkg/app/carbonapi/app.go
+++ b/pkg/app/carbonapi/app.go
@@ -45,6 +45,12 @@ type App struct {
 
 	backend backend.Backend
 
+	// During processing we use two independent queues that share a semaphore to prevent stampeding.
+	// fastQ includes regular requests
+	fastQ chan *renderReq
+	// slowQ contains large requests that could fill the queue and create a stampede.
+	slowQ chan *renderReq
+
 	ms PrometheusMetrics
 	Lg *zap.Logger
 
@@ -66,6 +72,8 @@ func New(config cfg.API, lg *zap.Logger, buildVersion string) (*App, error) {
 		defaultTimeZone: time.Local,
 		ms:              newPrometheusMetrics(config),
 		requestBlocker:  blocker.NewRequestBlocker(config.BlockHeaderFile, config.BlockHeaderUpdatePeriod, lg),
+		fastQ:           make(chan *renderReq, config.QueueSize),
+		slowQ:           make(chan *renderReq, config.QueueSize),
 	}
 	app.requestBlocker.ReloadRules()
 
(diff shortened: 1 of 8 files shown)


pkg/app/carbonapi/app.go
```diff
@@ -45,6 +45,12 @@ type App struct {
 
 	backend backend.Backend
 
+	// During processing we use two independent queues that share a semaphore to prevent stampeding.
+	// fastQ includes regular requests
+	fastQ chan *renderReq
+	// slowQ contains large requests that could fill the queue and create a stampede.
+	slowQ chan *renderReq
```
emadolsky, line 52: I think the naming can be improved.
slowQ makes me think of a queue in which requests are slow (like because they are big), or a queue which is read from less frequently.
But as I understood, slowQ is where a huge number of render requests which are sent to. And this huge number of render request are there because of glob breaking logic.
Surely, I don't see this as a blocker, but still I think the naming might cause some misunderstandings.

pkg/app/carbonapi/proc.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/bookingcom/carbonapi/pkg/carbonapipb"
+)
+
+// ProcessRequests processes the queued requests.
+// TODO: Handler request timeouts. The timed-out requests don't need to be forwarded.
+//       Currently, they'll be handled by the old limiter that's still in place.
+func ProcessRequests(app *App) {
+	// semaphore does what semaphores do: It limits the number of concurrent requests.
+	semaphore := make(chan bool, app.config.MaxConcurrentUpstreamRequests)
```
emadolsky, line 15: As this is only a semaphore, we can consider changing it to signal only channel by modifying it to `chan struct{}`

pkg/app/carbonapi/app_test.go
```diff
@@ -155,6 +155,9 @@ func SetUpTestConfig() (*App, http.Handler) {
 
 	setUpConfig(app, logger)
 	handler := initHandlers(app, logger)
+
+	go ProcessRequests(app)
```
emadolsky, line 159: I think this is not supposed to be run in a new goroutine although it'll work properly.

pkg/cfg/api.go
```diff
@@ -67,6 +67,26 @@ func DefaultAPIConfig() API {
...
+		MaxConcurrentUpstreamRequests: 20000,
+		// In most cases a single worker should suffice.
+		// The default is set to 4 as a precaution against bottlenecks.
+		ProcWorkers:  4,
+		LargeReqSize: 10000,
+
+		UpstreamSubRenderNumHistParams: HistogramConfig{
+			Start:      1,
+			BucketsNum: 12,
+			BucketSize: 3,
```
emadolsky, line 83: I think this histogram config can be tuned. Because subRender count is either 1 or > API.resolveGlobs.

pkg/cfg/api.go
```diff
@@ -78,26 +98,35 @@ func DefaultAPIConfig() API {
...
+
+	// The size of the requests queue propagated to backends.
+	// During this stage of refactoring it is a placeholder and should not fill-up.
+	// At the later stages it will play a key role in the request processing.
+	QueueSize                     int `yaml:"queueSize"`
+	MaxConcurrentUpstreamRequests int `yaml:"maxConcurrentUpstreamRequests"`
+	// The number of workers to process requests queue.
+	ProcWorkers int `yaml:"procWorkers"`
+	// The threshold of the number of sub-requests after which the render requests are considered large.
+	// It is used to select the processing queue: Small requests get on the fast queue, large ones on the slow one.
```
emadolsky, line 125: Queues consist of sub-requests. So, in order to make sure that the right message is transmitted, I propose:

> sub-requests derived from small requests get on the fast queue, and the ones derived from large requests get on the slow queue.

instead of:

> Small requests get on the fast queue, large ones on the slow one.

emadolsky (approved):
I left some comments. None of them are critical, so I also approved the PR.

pkg/app/carbonapi/proc.go
grzkv: I knew you'll have this comment) I did do this before, but I think bools are more clean.

pkg/app/carbonapi/app.go
grzkv: I get your point. I'd do the same 3 months ago.

However, the names must convey the semantics, not the processing speed. I think this is a better approach. I learned this from the Go runtime code where slow branches are named so not because they're slow per se, but semantically. This is a recent learning for me.

pkg/app/carbonapi/proc.go
grzkv: So, this is intentional. I always prefer bools to `struct{}` in such cases because `struct{}` has a more complicated semantics.

pkg/app/carbonapi/app_test.go
grzkv: Yes, you're right. This has to be corrected. I'll correct this in the next PR as it has no impact.

pkg/cfg/api.go
grzkv: I see your point, but I'd rather not complicate that. Also, I don't think it will play nice with the exponential buckets.

However, I can tell you for sure that we have to refactor the histogram configs in general. For example, we use the linear config for exponential histograms) So, that's definitely on the table.

pkg/cfg/api.go
grzkv: I'm not sure I see it that way. This is a matter of terminology.
I'll try to make things clearer moving forward.

PR #413, opened 2022-10-24
Processing redesign: Redesigned the backend requests processing.

grzkv:
*Disclaimer: This PR is part of a bigger refactoring. We had to make design compromises to accommodate both the old and the new processing systems at the same time. Once we completely switch to the new system, the design and code can be cleaned-up and ordered. Final touches are part of #407, and we'll do them before closing it.*

This is a continuation of #407

This PR contains the new implementation of the backend requests processing. The old backend limiters were removed and replaced with the new processing via queues. At the same time, the old top-level limiter in `carbonapi` still stays in place and will be removed at the later stages.

### DIFF
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -1,0 +1,173 @@
+package backend
+
+import (
+	"context"
+	"time"
+
+	"github.com/bookingcom/carbonapi/pkg/backend/mock"
+	"github.com/bookingcom/carbonapi/pkg/types"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+)
+
+type renderReq struct {
+	types.RenderRequest
+
+	Ctx       context.Context
+	StartTime time.Time
+
+	Results chan []types.Metric
+	Errors  chan error
+}
+
+type findReq struct {
+	types.FindRequest
+
+	Ctx       context.Context
+	StartTime time.Time
+
+	Results chan types.Matches
+	Errors  chan error
+}
+
+type infoReq struct {
+	types.InfoRequest
+
+	Ctx       context.Context
+	StartTime time.Time
+
+	Results chan []types.Info
+	Errors  chan error
+}
+
+// The backend to send requests to.
+type Backend struct {
+	BackendImpl
+
+	// Request queues.
+	// There's a separate one for each type, because the stampeding requests of one type should not
+	// impact the other types of requests.
+	// This, for example, removes the dependency between the find and render requests performance.
+	// Huge amount of render requests will not have direct impact on the find requests besides the semphore.
+	RenderQ chan *renderReq
+	FindQ   chan *findReq
+	InfoQ   chan *infoReq // the info requests are expected to be relatively rare
+
+	// The size of the requests semaphore:
+	// The maximum number of simultaneous requests.
+	SemaSize int
+}
+
+// Creates a new backend and starts processing the queues if qSize > 0.
+//
+// Passing qSize == 0 disables async processing, and the backend only works as a proxy to the implementation functions.
+// Using the channels won't work in that mode.
+// TODO (redesign): Remove the option to have qSize == 0 and proxy processor when the redesign is finished.
+func NewBackend(impl BackendImpl, qSize int, semaSize int,
+	requestsInQueue *prometheus.GaugeVec,
+	saturation *prometheus.Gauge,
+	timeInQSec *prometheus.HistogramVec) Backend {
+	b := Backend{
+		BackendImpl: impl,
+		RenderQ:     make(chan *renderReq, qSize),
+		FindQ:       make(chan *findReq, qSize),
+		InfoQ:       make(chan *infoReq, qSize),
+		SemaSize:    semaSize,
+	}
+
+	if qSize > 0 {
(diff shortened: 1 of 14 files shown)


pkg/backend/backend.go
```diff
@@ -0,0 +1,280 @@
...
+	enqueuedRequests *prometheus.CounterVec
+}
+
+// The specific backend implementation.
+//
+// At the moment of writing, it can be one of:
+// * mock, used for tests
+// * net, the standard protocol
+// * grpc, used for streaming gRPC
+type BackendImpl interface {
```
emadolsky, line 42: `BackendImpl` doesn't sound pretty well IMO as we have `Backend` as well. Maybe we can change it to `BackendQuerier` or `BackendConnector` for example.

pkg/backend/net/net.go
```diff
@@ -41,8 +41,8 @@ func (e ErrHTTPCode) Error() string {
 	}
 }
 
-// Backend represents a host that accepts requests for metrics over HTTP.
-type Backend struct {
+// NetBackend represents a host that accepts requests for metrics over HTTP.
+type NetBackend struct {
```
emadolsky, line 45: I suppose It makes sense to change this to `HTTPBackend`.

pkg/backend/backend.go
```diff
@@ -0,0 +1,280 @@
...
+				errCh <- err
+			} else {
+				msgCh <- msg
+			}
+		}(backend)
+	}
+}
+
+func NewMock(cfg mock.Config) Backend {
+	return NewBackend(mock.New(cfg), 0, 0, nil, nil, nil, nil)
```
emadolsky, line 279: Maybe it's worth adding `NewMockWithNewBackend` (or with a different name) initializing new backend using the new queues, and run the tests for this backend as well so that tests are also run against the new logic. It shouldn't be any different in the test logic, I guess.

emadolsky (changes requested):
Thanks for the big effort. I left some comments. Feel free to reply on them, or we can discuss them in a short call.

pkg/backend/backend.go
grzkv: I think the current name is the better than the suggestions, since it corresponds to a common naming pattern used in multiple projects. It also corresponds to the canonical design pattern and names things for what they are. I'm open to to alternatives, though.

pkg/backend/net/net.go
grzkv: I've considered that name. I discarded it, because it does not differentiate from the gRPC implementation which is also HTTP2.

pkg/backend/backend.go
grzkv: It is a typical anti-pattern in Go to repeat the package name in the symbols contained inside the package. If I understand correctly, this suggestion would do precisely that, duplicating the indication of a "backend".
If you mean something different, the old name still makes more sense to me, because it has no additional mocking besides the creation of the backend. 

Also, I hope to clean this up after the redesign is done. This is a necessary thing to keep the tests going.

grzkv: Also, let me kindly remind you of the disclaimer in the PR description) Such small things are not worth discussing on the current stage. A lot of the will change or go away.

pkg/backend/backend.go
```diff
@@ -0,0 +1,280 @@
...
+				<-semaphore
+				if b.saturation != nil {
+					b.saturation.Dec()
+				}
+			}(r)
+		}
+	}()
+	go func() {
+		for r := range b.findQ {
+			b.requestsInQueue.WithLabelValues("find").Dec()
```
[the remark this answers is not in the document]
grzkv: Good observation! However, it's the other way around, the check is already taken into account in the `SendXxx` functions. The other checks need to be removed. Let me do that.

pkg/backend/backend.go
```diff
@@ -0,0 +1,280 @@
...
+					req.Results <- res
+				}
+				<-semaphore
+				b.saturation.Dec()
+			}(r)
+		}
+	}()
+	go func() {
+		for r := range b.infoQ {
+			b.requestsInQueue.WithLabelValues("info").Dec()
```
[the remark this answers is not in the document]
grzkv: Good observation! However, it's the other way around, the check is already taken into account in the `SendXxx` functions. The other checks need to be removed. Let me do that.

pkg/backend/backend.go
```diff
@@ -0,0 +1,280 @@
...
+func (backend Backend) SendRender(ctx context.Context, request types.RenderRequest, msgCh chan []types.Metric, errCh chan error) {
+	if cap(backend.renderQ) > 0 {
+		backend.renderQ <- &renderReq{
+			RenderRequest: request,
+			Ctx:           ctx,
+			StartTime:     time.Now(),
+			Results:       msgCh,
+			Errors:        errCh,
+		}
+		backend.requestsInQueue.WithLabelValues("render").Inc()
```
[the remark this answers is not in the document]
grzkv: Good observation! However, it's the other way around, the check is already taken into account in the `SendXxx` functions. The other checks need to be removed. Let me do that.

grzkv:
@emadolsky Pushed the cleanup for the checks.

emadolsky (approved):

PR #414, opened 2022-10-26
Processing redesign: Disabled old limiter in carbonapi. Added timeout handling.

grzkv:
*Disclaimer: This PR is part of a bigger refactoring. We had to make design compromises to accommodate both the old and the new processing systems at the same time. Once we completely switch to the new system, the design and code can be cleaned-up and ordered. Final touches are part of #407, and we'll do them before closing it.*

This is a continuation of #407. The following changes were made:

* The old limiter is disabled in `carbonapi`;
* Timeout handling added to the top-level processing;
* Clean-up.

## Additional notes
* The old limiter is still in place to avoid huge PRs.
* The `find` requests are not limited on the top level to give them higher priority.
* Some benchmark were removed as non-indicative for the current system.

### DIFF
--- a/pkg/app/carbonapi/app.go
+++ b/pkg/app/carbonapi/app.go
@@ -77,7 +77,6 @@ func New(config cfg.API, lg *zap.Logger, buildVersion string) (*App, error) {
 	}
 	app.requestBlocker.ReloadRules()
 
-	// TODO(gmagnusson): Setup backends
 	backend, err := initBackend(app.config, lg,
 		app.ms.ActiveUpstreamRequests, app.ms.WaitingUpstreamRequests,
 		app.ms.UpstreamLimiterEnters, app.ms.UpstreamLimiterExits)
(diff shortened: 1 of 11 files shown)


auguzun (dismissed):

pkg/backend/benchmark_test.go
```diff
@@ -1,350 +0,0 @@
-package backend
```
deniszh, line 1: Why this test is deleted? Looks useful to me

grzkv: I don't think this measurement is in any way relevant at the moment. Tests like this introduce dependencies in the code and slow down the work while not bringing value.

The plan is to remove the majority of unit-tests for the backends as well as *all* http unit tests for zipper.

auguzun (dismissed):

deniszh (approved):

deniszh (approved):

PR #449, opened 2022-12-15
/metric/expand API implementation

deniszh:
Implementing [/metrics/expand API](https://graphite.readthedocs.io/en/latest/metrics_api.html#metrics-expand)

I added handler for expand route which call same `resolveGlobs()` as find does, but processing result differently using `expandEncoder` (similar as find uses `findEncoder` with different formats, but expand supports only json). Not sure it's a best approach, though, let's discuss that. Maybe we need to refactor it to decrease amount of copy-paste, but maybe we can do that in separate PR, this one is quite big even w/o that. 

I implemented test in http_handlers_test.go
Also I tested `expand()` function on local Grafana installation v8.15.3 and tested this on live traffic.

### DIFF
--- a/pkg/app/carbonapi/http_handlers.go
+++ b/pkg/app/carbonapi/http_handlers.go
@@ -917,7 +917,7 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
 	case treejsonFormat, jsonFormat:
 		blob, err = ourJson.FindEncoder(metrics)
 	case "", pickleFormat:
-		writeFormat = jsonFormat
+		writeFormat = pickleFormat
 		if app.config.GraphiteWeb09Compatibility {
 			blob, err = pickle.FindEncoderV0_9(metrics)
 		} else {
--- a/pkg/app/carbonapi/http_handlers.go
+++ b/pkg/app/carbonapi/http_handlers.go
@@ -1014,8 +1014,8 @@ func (app *App) expandHandler(w http.ResponseWriter, r *http.Request, lg *zap.Lo
 	app.ms.Requests.Inc()
 
 	query := r.FormValue("query")
-	leavesOnly := !parser.TruthyBool(r.FormValue("leavesOnly"))
-	groupByExpr := !parser.TruthyBool(r.FormValue("groupByExpr"))
+	leavesOnly := parser.TruthyBool(r.FormValue("leavesOnly"))
+	groupByExpr := parser.TruthyBool(r.FormValue("groupByExpr"))
 	jsonp := r.FormValue("jsonp")
 	useCache := !parser.TruthyBool(r.FormValue("noCache"))
 
--- a/pkg/app/carbonapi/http_handlers.go
+++ b/pkg/app/carbonapi/http_handlers.go
@@ -1037,8 +1037,8 @@ func expandEncoder(globs dataTypes.Matches, leavesOnly bool, groupByExpr bool) (
 		if leavesOnly && !g.IsLeaf {
 			continue
 		}
-		name := g.Path
-		nodes := strings.SplitN(name, ".", nodeCount+1)
+		var name string
+		nodes := strings.SplitN(g.Path, ".", nodeCount+1)
 		if len(nodes) > nodeCount {
 			name = strings.Join(nodes[:nodeCount], ".")
 		} else {
--- a/pkg/app/carbonapi/http_handlers.go
+++ b/pkg/app/carbonapi/http_handlers.go
@@ -959,14 +959,14 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
 	toLog.HttpCode = http.StatusOK
 }
 
-func expandEncoder(globs dataTypes.Matches, leavesOnly int, groupByExpr int) ([]byte, error) {
+func expandEncoder(globs dataTypes.Matches, leavesOnly bool, groupByExpr bool) ([]byte, error) {
 	var b bytes.Buffer
 	groups := make(map[string][]string)
 	seen := make(map[string]struct{})
 	nodeCount := len(strings.Split(globs.Name, "."))
 	names := make([]string, 0, len(globs.Matches))
 	for _, g := range globs.Matches {
-		if leavesOnly == 1 && !g.IsLeaf {
+		if leavesOnly && !g.IsLeaf {
 			continue
 		}
 		name := g.Path
@@ -981,13 +981,13 @@ func expandEncoder(globs dataTypes.Matches, leavesOnly int, groupByExpr int) ([]
 		}
 		seen[name] = struct{}{}
 		names = append(names, name)
-		sort.Strings(names)
 	}
+	sort.Strings(names)
 	groups[globs.Name] = names
 	data := map[string]interface{}{
 		"results": groups,
 	}
-	if groupByExpr != 1 {
+	if !groupByExpr {
 		flatData := make([]string, 0)
 		for _, group := range groups {
 			flatData = append(flatData, group...)
@@ -1014,22 +1014,13 @@ func (app *App) expandHandler(w http.ResponseWriter, r *http.Request, lg *zap.Lo
 	app.ms.Requests.Inc()
 
 	query := r.FormValue("query")
+	leavesOnly := !parser.TruthyBool(r.FormValue("leavesOnly"))
+	groupByExpr := !parser.TruthyBool(r.FormValue("groupByExpr"))
 	jsonp := r.FormValue("jsonp")
 	useCache := !parser.TruthyBool(r.FormValue("noCache"))
 
 	toLog := carbonapipb.NewAccessLogDetails(r, "expand", &app.config)
 	toLog.Targets = []string{query}
-
-	groupByExpr, err := strconv.Atoi(r.FormValue("groupByExpr"))
-	if (err != nil) || (groupByExpr != 0 && groupByExpr != 1) {
-		writeError(uuid, r, w, http.StatusBadRequest, "parameter `groupByExpr` should be either 0 or 1", "", &toLog)
-		return
-	}
-	leavesOnly, err := strconv.Atoi(r.FormValue("leavesOnly"))
-	if (err != nil) || (leavesOnly != 0 && leavesOnly != 1) {
-		writeError(uuid, r, w, http.StatusBadRequest, "parameter `leavesOnly` should be either 0 or 1", "", &toLog)
-		return
-	}
 
 	lg = lg.With(zap.String("request_id", uuid), zap.String("request_type", "expand"), zap.String("target", query))
 	Trace(lg, "received request")
--- a/pkg/app/carbonapi/http_handlers.go
+++ b/pkg/app/carbonapi/http_handlers.go
@@ -947,6 +947,82 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
 	toLog.HttpCode = http.StatusOK
 }
 
+func (app *App) expandHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logger) {
+	t0 := time.Now()
+	defer func() {
+		d := time.Since(t0).Seconds()
+		app.ms.DurationTotal.WithLabelValues("expand").Observe(d)
+	}()
+
+	ctx, cancel := context.WithTimeout(r.Context(), app.config.Timeouts.Global)
+	defer cancel()
+	uuid := util.GetUUID(ctx)
+
+	app.ms.Requests.Inc()
+
+	query := r.FormValue("query")
+	leavesOnly := parser.TruthyBool(r.FormValue("leavesOnly"))
+	groupByExpr := parser.TruthyBool(r.FormValue("groupByExpr"))
+	jsonp := r.FormValue("jsonp")
+	useCache := !parser.TruthyBool(r.FormValue("noCache"))
+
+	toLog := carbonapipb.NewAccessLogDetails(r, "expand", &app.config)
+	toLog.Targets = []string{query}
+
+	lg = lg.With(zap.String("request_id", uuid), zap.String("request_type", "expand"), zap.String("target", query))
+	Trace(lg, "received request")
+
+	logLevel := zap.InfoLevel
+	defer func() {
+		app.deferredAccessLogging(lg, r, &toLog, t0, logLevel)
+	}()
+
+	if query == "" {
+		writeError(uuid, r, w, http.StatusBadRequest, "missing parameter `query`", "", &toLog)
+		return
+	}
+	metrics, fromCache, err := app.resolveGlobs(ctx, query, useCache, &toLog, lg)
+	toLog.FromCache = fromCache
+	if err == nil {
+		toLog.TotalMetricCount = int64(len(metrics.Matches))
+	} else {
+		var notFound dataTypes.ErrNotFound
+
+		switch {
+		case errors.As(err, &notFound):
+			// we can generate 404 for expand, it's graphite-web-1.0 only
+			Trace(lg, "not found")
+			writeError(uuid, r, w, http.StatusNotFound, err.Error(), "", &toLog)
+			logLevel = zapcore.ErrorLevel
+			return
+		case errors.Is(err, context.DeadlineExceeded):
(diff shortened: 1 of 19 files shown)


pkg/app/carbonapi/http_handlers.go
```diff
@@ -944,39 +945,139 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
 		logLevel = zapcore.ErrorLevel
 		return
 	}
-
+	// TODO: extend writeResponse w/new formats and migrate to it
```
deniszh, line 948: Not needed in this change, I just found it very confusing that we have jsonp wrapper implemented in `writeResponse` already and find handler using separate wrapper for some reason.

pkg/app/carbonapi/http_handlers.go
```diff
@@ -944,39 +945,139 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
 			return
 		}
 	}
 
-	toLog.HttpCode = http.StatusOK
+	if ctx.Err() != nil {
+		app.ms.RequestCancel.WithLabelValues("expand", ctx.Err().Error()).Inc()
+	}
+
+	groups := make(map[string][]string)
```
deniszh, line 1040: As you can see above - everything from func expandhandler and up to this point is just copy of findHandler, but with different parameters, logging and metrics. Maybe we can simplify this, not sure if we should do that in the same PR, though.

pkg/types/encoding/json/encoding.go
```diff
@@ -56,9 +56,6 @@ func matchesToJSONMatches(matches types.Matches) []jsonMatch {
 			jm.Leaf = 1
 		} else {
 			jm.AllowChildren = 1
-		}
-
-		if !m.IsLeaf || strings.ContainsRune(jm.ID, '*') {
```
deniszh, line 61: Looks logical that we should enable Expandable for branches with globs - but that's not how graphite-web works. It sets expandable=true for all branches - see for reference https://github.com/graphite-project/graphite-web/blob/02bc0c836269e07b38175fbfe3b5689afc87b4fa/webapp/graphite/metrics/views.py#L315-L324 and https://github.com/graphite-project/graphite-web/blob/02bc0c836269e07b38175fbfe3b5689afc87b4fa/webapp/graphite/browser/views.py#L86-L95

pkg/app/carbonapi/http_handlers.go
```diff
@@ -939,39 +940,148 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
 			return
 		}
 	}
 
-	toLog.HttpCode = http.StatusOK
+	if ctx.Err() != nil {
+		app.ms.RequestCancel.WithLabelValues("expand", ctx.Err().Error()).Inc()
+	}
+
+	blob, err := expandEncoder(metrics, leavesOnly, groupByExpr)
```
deniszh, line 1074: As you can see above - almost everything in expandhandler is just a copy of findHandler, but with different parameters, logging and metrics. Maybe we can simplify this, not sure if we should do that in the same PR, though.

deniszh:
PS: implementation itself is borrowed from abandoned PR in https://github.com/go-graphite/carbonapi/pull/663

pkg/app/carbonapi/http_handlers.go
```diff
@@ -939,39 +940,148 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
-		if _, writeErr := w.Write([]byte{')'}); writeErr != nil {
-			toLog.HttpCode = 499
-			logLevel = zapcore.WarnLevel
-			return
+	}
+
+	toLog.HttpCode = http.StatusOK
+}
+
+func expandEncoder(globs dataTypes.Matches, leavesOnly string, groupByExpr string) ([]byte, error) {
```
emadolsky, line 962: I suggest changing `leavesOnly` to a boolean. the boolean value could be assigned to the boolean variable at the start of expandHandler.

emadolsky: Same for `groupByExpr`

deniszh: Why, if it's defined in API as a string? Not sure if I understand your suggestion clearly, though...

pkg/app/carbonapi/http_handlers.go
```diff
@@ -939,39 +940,148 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+			name = strings.Join(nodes[:nodeCount], ".")
+		} else {
+			name = strings.Join(nodes, ".")
+		}
+		if _, ok := seen[name]; ok {
+			continue
+		}
+		seen[name] = struct{}{}
+		names = append(names, name)
+		sort.Strings(names)
```
emadolsky, line 984: I think it might make more sense to sort the names after the loop.

pkg/app/carbonapi/http_handlers.go
deniszh: Ah, disregard that. It's defined as integer, probably will change it to int then.

pkg/app/carbonapi/http_handlers.go
emadolsky: Personally, I'd prefer to refactor findHandler first, so that it's easier and less verbose to implement expand. But you've already done it, and if we all agree that the refactoring here is crucial and we will do it afterwards, I think we can move forwards without refactoring at this moment.

grzkv:
@deniszh I believe a change like this one needs testing on live environment. Did we do it? (it does not state that in the description)

pkg/types/encoding/json/encoding.go
grzkv: Is this a separate change from adding `/metric/expenad`?

pkg/app/carbonapi/http_handlers.go
auguzun: I agree that the refactor needs to be done but maybe in separate task

auguzun (dismissed):

pkg/app/carbonapi/http_handlers.go
```diff
@@ -939,39 +940,148 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+	var b bytes.Buffer
+	groups := make(map[string][]string)
+	seen := make(map[string]struct{})
+	nodeCount := len(strings.Split(globs.Name, "."))
+	names := make([]string, 0, len(globs.Matches))
+	for _, g := range globs.Matches {
+		if leavesOnly == "1" && !g.IsLeaf {
+			continue
+		}
+		name := g.Path
```
timtofan, line 972: would it be more intuitive to call it `path := g.Path`?

pkg/app/carbonapi/http_handlers.go
deniszh: No, you're right bool is better. Please check.

pkg/app/carbonapi/http_handlers.go
deniszh: oops, good catch! fixed.

pkg/app/carbonapi/http_handlers.go
timtofan: agree w/ Emad, better to avoid duplication, so makes sense to refactor in same PR or as prev PR

but we can also merge as is and refactor later 🤷

pkg/app/carbonapi/http_handlers.go
deniszh: Sorry, but no. Too many changes.

deniszh: I mean then I need to rename names to paths, and then we'll get `groups[globs.Name] = paths`

deniszh:
@grzkv : sure, I tested it on DQS, but before merging I'll test it more ofc

pkg/app/carbonapi/http_handlers.go
deniszh: or I can do that in only 2 lines, if it's more pleasant to your eye, @timtofan :)
```
		path := g.Path
		nodes := strings.SplitN(path, ".", nodeCount+1)
```

deniszh: No, it cause more issues y=then it solves, sorry. :)
Let's keep it as is.

pkg/types/encoding/json/encoding.go
deniszh: Yes, it is. We have separate ticket for that. I can remove it if you hesitate.

pkg/app/carbonapi/http_handlers.go
deniszh: I'm not even sure that refactor is suitable here TBH. Yes, handlers look similar but I'm not sure where we can have comfortable level of duplication vs reimplementation of features.
Probably we should keep handlers for find and expand separate, but then we should have separate metrics and logging. What common left to refactor then?

pkg/types/encoding/json/encoding.go
deniszh: Probably, you're right, will exclude it and create separate PR

deniszh: Removed

deniszh:
I put expandable fix in separate PR https://github.com/bookingcom/carbonapi/pull/451

pkg/app/carbonapi/http_handlers.go
```diff
@@ -939,39 +940,148 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+	}()
+
+	ctx, cancel := context.WithTimeout(r.Context(), app.config.Timeouts.Global)
+	defer cancel()
+	uuid := util.GetUUID(ctx)
+
+	app.ms.Requests.Inc()
+
+	query := r.FormValue("query")
+	leavesOnly := !parser.TruthyBool(r.FormValue("leavesOnly"))
```
emadolsky, line 1017: I guess we should not have ! (not) on this line.

pkg/app/carbonapi/http_handlers.go
```diff
@@ -939,39 +940,148 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+	groupByExpr := !parser.TruthyBool(r.FormValue("groupByExpr"))
```
emadolsky, line 1018: I guess we should not have ! (not) on this line, as well.

deniszh: yes, good catch again, thanks, Emad!

deniszh:
@timtofan @emadolsky : OK, fine, I'll do refactor in this PR then, prolly make sense.

azhiltsov (dismissed):

pkg/app/carbonapi/http_handlers.go
```diff
@@ -940,38 +938,128 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+	blob, err := expandEncoder(metrics, leavesOnly, groupByExpr)
+	if err != nil {
+		writeError(uuid, r, w, http.StatusInternalServerError, err.Error(), "", &toLog)
+		logLevel = zapcore.ErrorLevel
+		return
+	}
+	if writeResponse(ctx, w, blob, jsonFormat, jsonp) != nil {
+		toLog.HttpCode = 499
+		logLevel = zapcore.ErrorLevel
+		return
```
grzkv, line 1022: I guess this return is not needed.

pkg/app/carbonapi/http_handlers.go
```diff
@@ -940,38 +938,128 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+	}
+}
+
+func expandEncoder(globs dataTypes.Matches, leavesOnly bool, groupByExpr bool) ([]byte, error) {
+	var b bytes.Buffer
+	groups := make(map[string][]string)
+	seen := make(map[string]struct{})
```
grzkv, line 1029: This would be more logical if it was a `map[string]bool`. It will not change the functionality, but will be more consistent.

auguzun (approved):

pkg/app/carbonapi/http_handlers.go
```diff
@@ -940,38 +938,128 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+	nodeCount := len(strings.Split(globs.Name, "."))
```
grzkv, line 1030: This is inefficient as it performs a string split with memory allocation for the resulting tokens. It introduces significant unnecessary memory allocations. We could simply count the number of `"."` instead. This way we don't allocate additional memory.

pkg/app/carbonapi/http_handlers.go
```diff
@@ -940,38 +938,128 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+		if leavesOnly && !g.IsLeaf {
+			continue
 		}
-	} else {
-		w.Header().Set("Content-Type", contentType)
-		if _, writeErr := w.Write(blob); writeErr != nil {
-			toLog.HttpCode = 499
-			logLevel = zapcore.WarnLevel
-			return
+		name := g.Path
```
grzkv, line 1036: This assignment looks semantically incorrect as it does not make sense to have `name == g.Path`. That value is _always_ overwritten. So, `name` is used as a proxy variable. It would make much more sense to have:
```suggestion
		nodes := strings.SplitN(g.Path, ".", nodeCount+1)
		var name string
```

pkg/app/carbonapi/http_handlers.go
```diff
@@ -940,38 +938,128 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+		nodes := strings.SplitN(name, ".", nodeCount+1)
```
grzkv, line 1037: This does a lot of unnecessary allocations. We split the string by `"."` and then join the resulting tokens by the same `"."`. Looks like we're getting a substring of the `g.Path`. To get a substring, we need only one allocation (actually, we could do 0, but that would be premature optimization), but we're doing much more than that.

pkg/app/carbonapi/http_handlers.go
deniszh: Yes, I forgot to set `toLog.HttpCode = http.StatusOK` later. Fixed.

pkg/app/carbonapi/http_handlers.go
```diff
@@ -940,38 +938,128 @@ func (app *App) findHandler(w http.ResponseWriter, r *http.Request, lg *zap.Logg
...
+			continue
 		}
+		seen[name] = struct{}{}
+		names = append(names, name)
 	}
-
-	toLog.HttpCode = http.StatusOK
+	sort.Strings(names)
+	groups[globs.Name] = names
+	data := map[string]interface{}{
```
grzkv, line 1051: I may be wrong, but I don't understand why this should be a `map[string]interface{}`. Looks like we can have an _if_ that does a separate encoding for both values of `groupByExpr`. We can then use specific data structures in both branches of _if_. I'm not sure if and how much it will improve performance (but I'd guess there will be a performance gain, because this way we will likely avoid an additional reflection during serialization), but it will certainly make the code much more clear.

pkg/app/carbonapi/http_handlers.go
deniszh: Agreed, fixed.

grzkv (approved):
@deniszh Thanks for taking into account my previous suggestions. I've left some comments, but they're not critical. We can address them later if it's more convenient for you.

pkg/app/carbonapi/http_handlers.go
deniszh: Good point. Fixed.

pkg/app/carbonapi/http_handlers.go
deniszh: Agreed. Fixed.

deniszh:
@grzkv : Good review, thanks! I addressed almost all your points, except last 2, which requires much more investigation and thinking, though.

auguzun (approved):

deniszh:
Thanks! Merging this for now, will think about further fixes later.

PR #467, opened 2023-01-20
Add config to respond 404 when TLD cache miss

emadolsky:
To avoid request fanout to all backends when a target is not in tld cache.

### DIFF
--- a/pkg/app/carbonapi/app.go
+++ b/pkg/app/carbonapi/app.go
@@ -41,10 +41,11 @@ type App struct {
 	config       cfg.API
 	ZipperConfig cfg.Zipper
 
-	queryCache             cache.BytesCache
-	findCache              cache.BytesCache
-	TopLevelDomainCache    *expirecache.Cache
-	TopLevelDomainPrefixes []tldcache.TopLevelDomainPrefix
+	queryCache               cache.BytesCache
+	findCache                cache.BytesCache
+	TopLevelDomainCache      *expirecache.Cache
+	TopLevelDomainPrefixes   []tldcache.TopLevelDomainPrefix
+	NotFoundWhenTLDCacheMiss bool
 
 	requestBlocker *blocker.RequestBlocker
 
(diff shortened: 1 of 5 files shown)


pkg/app/carbonapi/app.go
```diff
@@ -41,10 +41,11 @@ type App struct {
...
 
-	queryCache             cache.BytesCache
-	findCache              cache.BytesCache
-	TopLevelDomainCache    *expirecache.Cache
-	TopLevelDomainPrefixes []tldcache.TopLevelDomainPrefix
+	queryCache               cache.BytesCache
+	findCache                cache.BytesCache
+	TopLevelDomainCache      *expirecache.Cache
+	TopLevelDomainPrefixes   []tldcache.TopLevelDomainPrefix
+	NotFoundWhenTLDCacheMiss bool
```
timtofan, line 48: maybe NotFoundUponTLDCacheMiss? upon before noun, vs when before verf

deniszh (approved):

auguzun (approved):

pkg/app/carbonapi/app.go
timtofan: or maybe even better `ErrNotFoundUponTLDCacheMiss`

emadolsky: I'll keep this in mind for the next times and refactor this. thx.

pkg/tldcache/tldcache.go
```diff
@@ -134,23 +134,37 @@ func getTargetTopLevelDomain(target string, prefixes []TopLevelDomainPrefix) str
...
 	bs := filterByTopLevelDomain(cache, backends, targetTlds)
 	if len(bs) > 0 {
-		return bs
+		return bs, false
+	}
+	tldCacheMiss := true
+	for _, targetTLD := range targetTlds {
+		if strings.ContainsAny(targetTLD, "*?[{") {
+			tldCacheMiss = false
+			break
```
timtofan, line 164: maybe `return backends, true`?

PR #468, opened 2023-01-20
Verbose error message for moving window functions

auguzun:
for moving window functions there will be verbose error if the window size is bigger than the metrics retention

### DIFF
--- a/pkg/parser/interface.go
+++ b/pkg/parser/interface.go
@@ -65,6 +65,8 @@ var (
 	ErrDifferentCountMetrics = ParseError("both arguments must have the same number of metrics")
 	// ErrInvalidArgumentValue is an eval error returned when a function received an argument that has the right type but invalid value
 	ErrInvalidArgumentValue = ParseError("invalid function argument value")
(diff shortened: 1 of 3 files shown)


pkg/parser/interface.go
```diff
@@ -65,6 +65,8 @@ var (
 	ErrDifferentCountMetrics = ParseError("both arguments must have the same number of metrics")
 	// ErrInvalidArgumentValue is an eval error returned when a function received an argument that has the right type but invalid value
 	ErrInvalidArgumentValue = ParseError("invalid function argument value")
+	// ErrMovingWindowSizeLessThanRetention is an eval error returned when a function received a moving window size bigger than the metrics retention
+	ErrMovingWindowSizeLessThanRetention = ParseError("moving window is bigger than the metrics retention")
```
timtofan, line 69: I'd rephrase to `windowSize param exceed metric retention`

auguzun: thanks, fixed

azhiltsov (approved):

PR #473, opened 2023-02-01
Add source clusters to render logs

emadolsky:
This PR adds a SourceClusters to metrics to identify the cluster the metric originated from. When merging the metrics from multiple upstreams, the result metric contains all of the backends involved for that metric in its SourceClusters. In the end, carbonapi, logs the request with all of the source clusters who contributed for gathering the metrics.

### DIFF
--- a/pkg/app/carbonapi/http_handlers.go
+++ b/pkg/app/carbonapi/http_handlers.go
@@ -303,6 +303,7 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request, lg *zap.Lo
 
 		Trace(lgt, "target succeeded")
 	}
+	toLog.Clusters = clustersFromMetricMap(metricMap)
 	toLog.CarbonzipperResponseSizeBytes = int64(size * 8)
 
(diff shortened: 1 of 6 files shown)


pkg/app/carbonapi/http_handlers.go
```diff
@@ -338,6 +339,23 @@ func (app *App) renderHandler(w http.ResponseWriter, r *http.Request, lg *zap.Lo
 	}
 }
 
+func clustersFromMetricMap(metricMap map[parser.MetricRequest][]*types.MetricData) []string {
```
deniszh, line 342: That looks like quite heavy function, and we're calling it for every metric. Maybe we can cache/memoize result of it?

grzkv (approved):

auguzun (approved):

pkg/app/carbonapi/http_handlers.go
deniszh: Ok, discussed offline, maybe it's excessive.

deniszh (approved):

PR #478, opened 2023-02-16
fix panic because integer overflow in valuesPerPoint

deniszh:
```
http: panic serving 127.0.0.1:44020: runtime error: slice bounds out of range [:-9223372036854775808]
goroutine 1134454440 [running]:
net/http.(*conn).serve.func1()
        /usr/lib64/go/src/net/http/server.go:1850 +0xbf
panic({0xd4bea0, 0xc3c0dc5b00})
        /usr/lib64/go/src/runtime/panic.go:890 +0x262
github.com/bookingcom/carbonapi/pkg/expr/types.(*MetricData).Consolidate(0xc076caf520, 0x8000000000000000)
        /usr/local/git_tree/gopath/src/github.com/bookingcom/carbonapi/pkg/expr/types/types.go:295 +0x6f0
github.com/bookingcom/carbonapi/pkg/expr/types.ConsolidateJSON(0x42b, {0xc1c75fbf00, 0xf, 0x2000100?})
        /usr/local/git_tree/gopath/src/github.com/bookingcom/carbonapi/pkg/expr/types/types.go:130 +0x190
github.com/bookingcom/carbonapi/pkg/app/carbonapi.(*App).renderWriteBody(0xc00025b800?, {0xc1c75fbf00?, 0xf, 0x10}, {{0xc35e47bf20, 0x3, 0x3}, {0xc066986052, 0xa}, {0xc066986063>
        /usr/local/git_tree/gopath/src/github.com/bookingcom/carbonapi/pkg/app/carbonapi/http_handlers.go:708 +0xcb
github.com/bookingcom/carbonapi/pkg/app/carbonapi.(*App).renderHandler(0xc00025b800, {0xee25b8, 0xc41a0dcf80}, 0xc304c82100, 0x1651240?)
```

valuesPerPoint is calculated from maxDataPoints and from/until parameters. It is defined as `int` in many places, redefining or converting to uint would be quite hard. 
So, first, I denying maxDataPoints to be < 0, which is kinda logical, negative values was never supported, and we pretty much don't care about int overflow there either. 
Then I just assuming that if valuesPerPoint is negative (either because of int overflow or some other error in data) we can treat that case the same as valuesPerPoint is 0 or 1, i.e. not do any aggregation.
Not sure if best approach, though.

Tested on prod, didn't find any issues.

### DIFF
--- a/pkg/expr/types/types.go
+++ b/pkg/expr/types/types.go
@@ -269,7 +269,7 @@ func MarshalRaw(results []*MetricData) []byte {
 // Consolidate returns a consolidated copy of this MetricData.
 func (r *MetricData) Consolidate(valuesPerPoint int) *MetricData {
 	ret := *r
-	if valuesPerPoint == 1 || valuesPerPoint == 0 {
+	if valuesPerPoint == 1 || valuesPerPoint <= 0 {
 		ret.ValuesPerPoint = 1
 		ret.Values = make([]float64, len(r.Values))
 		ret.IsAbsent = make([]bool, len(r.IsAbsent))
--- a/pkg/app/carbonapi/http_handlers.go
+++ b/pkg/app/carbonapi/http_handlers.go
@@ -704,7 +704,7 @@ func (app *App) renderWriteBody(results []*types.MetricData, form renderForm, r 
 
 	switch form.format {
 	case jsonFormat:
-		if maxDataPoints, _ := strconv.Atoi(r.FormValue("maxDataPoints")); maxDataPoints != 0 {
+		if maxDataPoints, _ := strconv.Atoi(r.FormValue("maxDataPoints")); maxDataPoints > 0 {
 			results = types.ConsolidateJSON(maxDataPoints, results)
 		}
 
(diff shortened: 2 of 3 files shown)


pkg/expr/types/types.go
```diff
@@ -269,8 +269,8 @@ func MarshalRaw(results []*MetricData) []byte {
 // Consolidate returns a consolidated copy of this MetricData.
 func (r *MetricData) Consolidate(valuesPerPoint int) *MetricData {
 	ret := *r
-	if valuesPerPoint == 1 || valuesPerPoint == 0 {
-		ret.ValuesPerPoint = 1
+	if valuesPerPoint == 1 || valuesPerPoint <= 0 {
+		ret.ValuesPerPoint = valuesPerPoint
```
emadolsky, line 273: Could you please elaborate why this line is changed?
If I understand correctly, this means that if `valuesPerPoint` is negative, put it in `ret.ValuesPerPoint` which doesn't seem correct. Also, I think if the panic observed is because of `maxDataPoints` being negative, we'll never reach this line because of the new change with `maxDataPoints > 0` condition on http_handlers.go file, right?

emadolsky (changes requested):

pkg/expr/types/types.go
deniszh: Please check my latest commit, @emadolsky

emadolsky (approved):

PR #483, opened 2023-03-03
Added check for the image size before calling cairo.

grzkv:
Looks like `cairo.ImageSurfaceCreate` can be called with the image size larger than the one allowed.
The maximum can be looked up e.g. here https://github.com/ImageMagick/cairo/blob/main/src/cairo-image-surface.c#L62
It's also mentioned in the mailing lists https://cairo.cairographics.narkive.com/Mkjwvkep/32k-limit-with-image-surface The lists are old, but seems like the limitation is still in place.

Added a check.


timtofan (approved):

deniszh (approved):
I did similar fix, it probably can be removed after this

PR #521, opened 2024-08-20
Fix integration test on latest ubuntu

deniszh:


deniszh:
closed in the favour of #522

PR #534, opened 2024-11-30
fix SummarizeValues

### LINKED ISSUE #533: Issues with legendValue
I want to display performance values within the graphs legend, the config looks like this:
`
legendValue(alias(color($metric$, '#5ff442'), 'In'), 'last', 'avg', 'max', 'si')
`

The issue is that 'last' and 'average' are displayed correctly after new data has been written, but after a certain time, 'last' always shows 0 and 'average' does not seem to be calculated correctly either.
This cycle repeats when new data is written.

The 'max' value is always displayed correctly.

legend
legend-issue

spacefreak86:
This PR fixes #533 (wrong last value being shown in legend) and also corrects the calculation of the average, min, last and count values.

By adding the parameter `absent` to `SummarizeValues`, it is now able to correctly account for absent values.

Logically all calls to SummarizeValues also had to be adapted.

### DIFF
--- a/pkg/expr/functions/legendValue/function.go
+++ b/pkg/expr/functions/legendValue/function.go
@@ -57,7 +57,7 @@ func (f *legendValue) Do(ctx context.Context, e parser.Expr, from, until int32, 
 	for _, a := range arg {
 		var values []string
 		for _, method := range methods {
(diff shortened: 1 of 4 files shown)


deniszh:
@grzkv @emadolsky : could you please also review?

deniszh (approved):

PR #553, opened 2026-02-06
Support none type

### LINKED ISSUE #295: parser does not support explicit None arguments for functions
For `asPercent`:
https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.asPercent
we can have `total1 set to None explicitly when we also have nodes.

Anandbrewery:
PR is dedicated to resolve None type issue faced below
Fixes [#295](https://github.com/bookingcom/carbonapi/issues/295)

- EtNil constant is introduced in interface.go 
- By intercepting "None" (case-insensitive) in parser.go and assigning it EtNil, we provide a type-safe way for functions to handle  null arguments. This follows the existing architecture for constants and strings.

- Added unit tests in parser_test.go for standalone None and nested function calls.
- All 46 TestParseExpr cases passed on running-> `(go test -v ./pkg/parser/... -run TestParseExpr)`

### DIFF
--- a/pkg/parser/interface.go
+++ b/pkg/parser/interface.go
@@ -24,6 +24,8 @@ const (
 	EtConst
 	// EtString is a const for 'String' type expression
 	EtString
+	// EtNil is a const for 'None' type expression
+	EtNil
 )
 
 var (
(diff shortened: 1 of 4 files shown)


Anandbrewery:
hi @deniszh @grzkv @azhiltsov @timtofan @emadolsky  should i submit this PR with single commit or multi commit for each change is fine.

Anandbrewery:
hi @deniszh @grzkv could please let me know is there any issue with this PR

Issue #503, opened 2024-03-28
Can't get internal metrics from carbonapi and carbonzipper


BorisMan9:
in my config yaml (both for carbonapi and carbonzipper) I configured the internal metrics prop:
```
graphite:
  host: "localhost:2003"
  interval: "60s"
  prefix: "carbon.api" // ("carbon.zipper" for carbonzipper)
  pattern: "{prefix}.{fqdn}"
```

but the internal metrics are not created

PR #555, opened 2026-05-25
fix(http): return structured JSON errors from render endpoint

m-zandi:
## Summary

Improves error handling across the HTTP render and metrics/find endpoints
by introducing structured JSON error responses and typed sentinel errors.

## Problem

When one or more upstream backends fail, carbonapi currently:

- Returns HTTP 200 with empty data (silent failure)
- Uses unstructured plain-text errors via `http.Error()`
- Provides no way for clients to distinguish backend failures
  from legitimately empty queries
- Drops request context (no request ID in errors)

This makes debugging production incidents extremely difficult,
as dashboards show "no data" with no indication of cause.
Related to issue #503.

## Solution

1. Added structured JSON error responses inside `writeError()`
2. Updated render handler to return HTTP 503 on full backend failure
3. Updated render handler to return HTTP 400 on bad requests
4. Propagates `X-Request-ID` header into all error responses
5. Added unit tests covering JSON serialization and edge cases

## Example Response After This Change

```json
{
  "error": "backends unavailable",
  "status": 503,
  "request_id": "1a2b3c4d"
}
```

## Tests

- [x] Unit tests for JSON error serialization
- [x] Unit tests for empty `request_id` omission
- [x] `go test -race ./pkg/types/...` passes
- [x] `go build ./...` clean
- [x] Existing test suite unchanged (no regressions)

## Breaking Changes

Error response format changes from plain text to JSON.
HTTP status codes on backend failures change from 200 to 503.
Clients relying on HTTP 200 for all responses need updating.
This is intentional — the old behavior was a bug.

## Notes for Reviewers

- Happy to split into smaller commits if preferred
- Open to feedback on the JSON field names
- The `writeError()` approach follows the existing pattern
  already used in the render and find handlers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ coverage.txt
 *.swp
 # Docker runs data
 data
+
+golangci-lint.zip
+golangci-lint-extracted/
--- a/pkg/app/carbonapi/http_handlers.go
+++ b/pkg/app/carbonapi/http_handlers.go
@@ -32,7 +32,10 @@ import (
 	ourJson "github.com/bookingcom/carbonapi/pkg/types/encoding/json"
 	"github.com/bookingcom/carbonapi/pkg/types/encoding/pickle"
 	"github.com/bookingcom/carbonapi/pkg/util"
+
 )
+
+
 
 const (
 	jsonFormat      = "json"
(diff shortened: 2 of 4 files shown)


deniszh:
Hi @m-zandi ,
Thanks for your contribution but we can't merge it to upstream as is. We're not treating old behavior as a bug but vice versa, despite it against common agreement for frontend errors. Also, our production code relies on current error scheme.

